PR #1, opened 2019-01-05
Add inactive imagestate

bear454:
Long-term supported images are neither actively maintained, nor in the 6-month deprecation window, so we need a new state: inactive.

RE: https://trello.com/c/KXt5oOZ9


ikapelyukhin:
There's another list of states in the readme that's missing `inactive`:

https://github.com/SUSE-Enceladus/public-cloud-info-service/blob/2ff0a9c8520744deada248dcccae0619883d81de/README.md#L7-L8

The existing image states seem to be covered with specs/fixtures a bit more thoroughly:

https://github.com/SUSE-Enceladus/public-cloud-info-service/blob/2ff0a9c8520744deada248dcccae0619883d81de/spec/features/app_spec.rb#L286-L304

bear454:
Thanks @ikapelyukhin ; resolved in 51a06b7, 7443eb3.

bear454:
Rebased onto master branch; squashed fix commits.

ikapelyukhin:
Just an observation: `publishedon="20180815" deprecatedon="" deletedon=""` -- there are timestamps for image state changes for the old states. Perhaps there also should be `deactivatedon`?

bear454:
> Just an observation: `publishedon="20180815" deprecatedon="" deletedon=""` -- there are timestamps for image state changes for the old states. Perhaps there also should be `deactivatedon`?

I don't think we need it; active & inactive differentiate how we maintain the images, not whether or not we do. Deprecated clearly indicates something is on its way out, and deleted is just historic data (e.g. "Where did that AMI go?")... so I thought about it, but didn't feel it was necessary.

I'll leave it to @rjschwei to decide if he thinks it's important.

rjschwei:
deprecatedon has a distinct meaning, as in "the image will go away in 6 months". The "inactive" state has no such clear definition, it only means "no more refreshes for this image" Thus a "deactivatedon" would be another setting for historical purposes. I think we can do without it. If for some reason this proves to be relevant at some point later the data can easily be added by using lifecycle information as the date would coincide with the beginning of LTSS fro SLES or SLES For SAP

PR #5, opened 2020-08-11
Add metadata requests to pint API (CSE-20)

chajain:
Added APIs to allow users to query for available metadata

/v1/providers.?ext? - user can query Pint via REST api consistent with existing queries for a list of known providers/frameworks
/v1/:provider/regions.?ext? - user can query Pint via REST api consistent with existing queries for a list of known regions
/v1/:provider/images/states.?.ext? - user can query Pint via REST api consistent with existing queries for a list of states
/v1/:provider/servers/types.?ext? - user can query Pint via REST api consistent with existing queries for a list of server types

Added the tests for the new APIs
Modified README.md to add new APIs and also instructions to setup Development environment

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,28 @@ all information at that level.
 
 For example:
 
+**Metadata APIs**
+```
+https://susepubliccloudinfo.suse.com/v1/providers.json
+```
+Will return list of all providers in JSON format 
+
+```
+https://susepubliccloudinfo.suse.com/v1/images/states.json
+```
+Will return list of all image states in JSON format
+
+```
+https://susepubliccloudinfo.suse.com/v1/amazon/servers/types.json
+```
+Will return list of all server types in Amazon in JSON format
+
+```
+https://susepubliccloudinfo.suse.com/v1/amazon/regions.json
+```
+Will return list of all known regions in Amazon in JSON format
+
+**Specific APIs**
 ```
 https://susepubliccloudinfo.suse.com/v1/microsoft/West%20US/servers/smt.json
 ```
@@ -113,3 +135,40 @@ Client Implementation
 
 `pint`, our Python-based client, is available in the Public Cloud Module, or in
 the Cloud:Tools project in OBS, as `python-susepubliccloudinfo`.
+
+Development
+-----------
+The following steps are only recommended if you want to build this project from sources, work on the codebase or test the latest development changes
+
+  1. **Install Git**
+
+         $ sudo zypper in git
+
+  2. **Install basic Ruby environment**
+
+         $ sudo zypper in ruby
+
+  3. **Install dependencies**
+
+     Install packages needed to compile Gems with native extensions:
+
+         $ sudo zypper in gcc make ruby-devel ruby2.5-rubygem-rack
+
+  4. **Clone public-cloud-info-service repository and install Gem dependencies**
+
+         $ git clone https://github.com/SUSE-Enceladus/public-cloud-info-service.git 
+         $ cd public-cloud-info-service 
+         $ bundle install
+
+  6. **In order to run REST APIs on the host**
+        
+         $ cd public-cloud-info-service
+         $ export FRAMEWORKS=spec/fixtures/framework*.xml
+         $ rackup
+         
+      To run REST APIs, point your browser to http://127.0.0.1:9292/
+
+  7. **In order to run tests**
+   
+         $ cd public-cloud-info-service
+         $ rspec
--- a/app.rb
+++ b/app.rb
@@ -175,6 +175,50 @@ class PublicCloudInfoSrv < Sinatra::Base
     end
   end
 
+  def form_xml_for_settings(tag, values)
+    Nokogiri::XML::Builder.new do |xml|
+      xml.root do
+        values.each { |value| xml.send(tag, name: value) }
+      end
+    end.doc.css(tag)
+  end
+
+  get '/v1/providers.?:ext?' do
+    validate_params_ext
+
+    responses = form_xml_for_settings('provider', settings.providers)
+
+    respond_with params[:ext], :providers, responses
+  end
+
+  get '/v1/:provider/servers/types.?:ext?' do
+    validate_params_ext
+    validate_params_provider
+
+    responses = form_xml_for_settings('type', settings.server_types)
+
+    respond_with params[:ext], :types, responses
+  end
+
+  get '/v1/images/states.?:ext?' do
+    validate_params_ext
+
+    responses = form_xml_for_settings('state', settings.image_states)
+
+    respond_with params[:ext], :states, responses
+  end
+
+  get '/v1/:provider/regions.?:ext?' do
+    validate_params_ext
+    validate_params_provider
+
+    responses = form_xml_for_settings(
+      'region', settings.regions[params[:provider]]
+    )
+
+    respond_with params[:ext], :regions, responses
+  end
+
   get '/v1/:provider/:region/servers/:server_type.?:ext?' do
     validate_params_ext
     validate_params_server_type
--- a/spec/features/app_spec.rb
+++ b/spec/features/app_spec.rb
@@ -513,3 +513,96 @@ describe 'environmented images: /v1/microsoft' do
     end
   end
 end
+
+# we test the metadata APIs for amazon
+describe 'metadata APIs' do
+  describe 'providers list' do
+    before do
+      @path = '/v1/providers.xml'
+    end
+
+    it 'responds successfully' do
+      get 'v1/providers'
+      expect(last_response.status).to eq 200
+    end
+
+    it 'matches Json provider list' do
+      get 'v1/providers'
+      expect(last_response.body).to eq(
+        '{"providers":[{"name":"amazon"},{"name":"google"},' \
+        '{"name":"microsoft"},{"name":"oracle"}]}'
+      )
+    end
+
+    it 'matches Xml provider list' do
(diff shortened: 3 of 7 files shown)


jeremy-moffitt:
@bear454 it doesnt appear that Charu or I can add reviewers to this, please take a look when you have time.

guangyee:
Code looks good. I presume we'll need to update API Doc somewhere?

bear454:
> Code looks good. I presume we'll need to update API Doc somewhere?

Please document API changes in the README.

bear454:
Sorry, I've made some changes to the application, you'll have to rebase on master and resolve conflicts. This is the first time the project really has contributors, so I've added style checks in `rubocop` and automated the test suite via `travis-ci`.

Please also extend the test suite to cover your use cases.

app.rb
```diff
@@ -175,6 +175,69 @@ def respond_with(format, category, responses)
...
+
+    responses = settings_xml.doc.css("servers>server")
+
+    respond_with params[:ext], :servers, responses
+  end
+
+  get '/v1/:provider/images/states.?:ext?' do
+    validate_params_ext
+    validate_params_provider    
+
```
rjschwei, line 222: This should not really be a dynamic implementation. The life-cycle has 4 defined states: `active`, `inactive`, `deprecated`, and `deleted`. Whether or not we have `deleted` images in any given framework is immaterial. I agree that a user should have to implicitly know the states thus a query API for providing the defined states is good. But this should be framework independent, i.e. `/v1/images/states.?:ext?` and the should return the static list of the defined states, no data processing needed.

spec/features/app_spec.rb
```diff
@@ -510,3 +510,34 @@
     end
   end
 end
+
+# we test the metadata APIs
+describe 'metadata APIs' do
+  describe 'providers list' do
+    it 'responds successfully' do
+      get "v1/providers"
+      expect(last_response.status).to eq 200
```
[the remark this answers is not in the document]
rjschwei: Yes, we need to check what is returned.

spec/features/app_spec.rb
```diff
@@ -513,3 +513,34 @@
...
+    end
+  end
+
+  describe 'server types list' do
+    it 'responds successfully' do
+      get "v1/#{$valid_providers.first}/servers/types"
+      expect(last_response.status).to eq 200
```
rjschwei, line 529: Again, the return value is important, not just that the server responds with 200.

spec/features/app_spec.rb
```diff
@@ -513,3 +513,34 @@
...
+    end
+  end
+  
+  describe 'image states' do
+    it 'responds successfully' do
+      get "v1/#{$valid_providers.first}/images/states"
+      expect(last_response.status).to eq 200
```
rjschwei, line 536: Ditto

spec/features/app_spec.rb
```diff
@@ -513,3 +513,34 @@
...
+    end
+  end
+
+  describe 'regions list' do
+    it 'responds successfully' do
+      get "v1/#{$valid_providers.first}/regions"
+      expect(last_response.status).to eq 200
```
rjschwei, line 543: And one more time.

app.rb
bear454: Well, we do need to check that `:ext` is valid, but I agree it shouldn't be provider-specific.

README.md
```diff
@@ -113,3 +135,44 @@ Client Implementation
...
+
+  4. **Clone public-cloud-info-service repository and install Gem dependencies**
+
+         $ git clone https://github.com/SUSE-Enceladus/public-cloud-info-service.git 
+         $ cd public-cloud-info-service 
+         $ bundle install
+
+  5. **Clone pint-data repository to import the xml files**
+
+         $ git clone gitlab@gitlab.suse.de:public-cloud-team/pint-data.git
```
bear454, line 165: Alternatively, you can use the included `spec/fixtures` data; e.g.

`FRAMEWORKS=spec/fixtures/frameworks.xml rackup`

bear454: Since the gitlab data is private, and this project is public, it would be better to have the README reflect the public data sample.

README.md
```diff
@@ -113,3 +135,44 @@ Client Implementation
...
+         $ export FRAMEWORKS=~/pint-data/data/*.xml
+         $ cd public-cloud-info-service
+         $ rackup
+         
+      To run REST APIs, point your browser to https://127.0.0.1:9292/
+
+  7. **In order to run tests**
+   
+         $ cd public-cloud-info-service
+         $ rake
```
bear454, line 178: `rspec`

app.rb
```diff
@@ -175,6 +175,69 @@ def respond_with(format, category, responses)
...
+      }
+    end
+    return builder
+  end
+
+  get '/v1/providers.?:ext?' do
+    validate_params_ext
+    
+    settings_xml = form_xml_for_settings(:providers, :name, settings.providers)
+    responses = settings_xml.doc.css("providers>provider")
```
bear454, line 202: Instead of sending the parent tag, using a case to set the child tag, then filtering out the children, please consider explicitly sending the child tag, and in the processor, setting a generic parent, and filtering it off; e.g.

```
  def form_xml_for_settings(tag, values)
    Nokogiri::XML::Builder.new do |xml|
      xml.root {
        values.each { |value| xml.send(tag, value) }
      }
    end.doc.css(tag)
  end

  get '/v1/states.?:ext?' do
    validate_params_ext
    responses = form_xml_for_settings('state', settings.image_states)
    respond_with params[:ext], :states, responses
  end
```

README.md
rjschwei: Yes, I completely missed that. There cannot be any references to gitlab.suse.de here. This very pull request leaks internal to SUSE data, grmbl. We should know better.

README.md
```diff
@@ -113,3 +135,40 @@ Client Implementation
...
+-----------
+The following steps are only recommended if you want to build this project from sources, work on the codebase or test the latest development changes
+
+  1. **Install Git**
+
+         $ sudo zypper in git
+
+  2. **Install basic Ruby environment**
+
+         $ sudo zypper in ruby rubygem-bundler
```
jeremy-moffitt, line 149: as an FYI, rubygem-bundler wasnt found for me on 15.2 , though "bundle" below worked find regardless

README.md
```diff
@@ -113,3 +135,40 @@ Client Implementation
...
+         $ cd public-cloud-info-service 
+         $ bundle install
+
+  6. **In order to run REST APIs on the host**
+        
+         $ cd public-cloud-info-service
+         $ export FRAMEWORKS=spec/fixtures/framework*.xml
+         $ rackup
+         
+      To run REST APIs, point your browser to https://127.0.0.1:9292/
```
jeremy-moffitt, line 169: https gets me ssl errors consistently, guessing this isnt setup with a certificate for the local environment... works fine with plain http

app.rb
```diff
@@ -175,6 +175,50 @@ def respond_with(format, category, responses)
...
+      xml.root do
+        values.each { |value| xml.send(tag, name: value) }
+      end
+    end.doc.css(tag)
+  end
+
+  get '/v1/providers.?:ext?' do
+    validate_params_ext
+
+    responses = form_xml_for_settings('provider', settings.providers)
```
rjschwei, line 189: We should come back with another PR and make this dynamic driven based off the .xml file names in the data location. Then it will be possible to onboard a new provider by simply dropping the data file into the "right" location.

app.rb
```diff
@@ -175,6 +175,50 @@ def respond_with(format, category, responses)
...
+
+  get '/v1/:provider/regions.?:ext?' do
+    validate_params_ext
+    validate_params_provider
+
+    responses = form_xml_for_settings(
+      'region', settings.regions[params[:provider]]
+    )
+
+    respond_with params[:ext], :regions, responses
```
rjschwei, line 219: This has to be revisited considering https://bugzilla.suse.com/show_bug.cgi?id=1169180. In Azure there are multiple names that need to get mapped, thus the simple loop over the data may not be sufficient. We have to have a conversation about this. On the other hand we may not care to expose all possible values through this API, in which case the current implementation is sufficient.

README.md
bear454: No, sinatra does not inherently process TLS; this has to be handled though an additonal module or (most commonly) via a reverse proxy.

app.rb
bear454: :thinking: At the rate we're currently adding providers, I'm happy with maintaining the list here as a validator. If we start adding providers with a frequency that makes the list difficult to maintain, then we can consider that change.

app.rb
bear454: Yes, we need to extend the 'region list' design that's currently in Azure into the other providers, and extend it to have a canonical region name, with acceptable aliases. This is going to be a pretty big rework project :/

spec/features/app_spec.rb
```diff
@@ -513,3 +513,64 @@
     end
   end
 end
+
+# we test the metadata APIs for amazon
+describe 'metadata APIs' do
```
bear454, line 518: Since there was an implementation question about the JSON responses, please add some tests on the new endpoints explicitly testing both XML and JSON formats, to prove the results are equivalent. Thanks.

bear454: After that, it looks good to me.

bear454:
@chajain anything else to add, or can I merge this and bump the version?

chajain:
Please merge

Thanks,
Charu


________________________________

PR #6, opened 2020-08-12
Fix DocumentRoot path

guangyee:
It should be /srv/www/publicCloudInfo-server/public instead of
/srv/www/publicCloudInfoSrv/public, according to
https://github.com/SUSE-Enceladus/public-cloud-info-service/blob/master/publicCloudInfo-server.spec#L55

### DIFF
--- a/publicCloudInfo-server.conf.template
+++ b/publicCloudInfo-server.conf.template
@@ -1,13 +1,13 @@
 <VirtualHost *:80>
-  ServerAdmin jgordon@suse.com
+  ServerAdmin admin@example.com
   ServerName susepubliccloudinfo.suse.com
-  DocumentRoot /srv/www/publicCloudInfoSrv/public
+  DocumentRoot /srv/www/publicCloudInfo-server/public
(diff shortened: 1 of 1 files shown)


publicCloudInfo-server.conf.template
```diff
@@ -1,13 +1,13 @@
 <VirtualHost *:80>
   ServerAdmin jgordon@suse.com
```
ikapelyukhin, line 2: We might want to remove @joelgordon's email from here :slightly_smiling_face:

joelgordon: It can be changed to publiccloud-admin@susecloud.net

guangyee: Which email should we use here? Or should we just remove the ServerAdmin config?

ikapelyukhin: @guangyee well, it's a template, so I'd go for some dummy address like `email@example.org`.

PR #7, opened 2020-08-27
Enable CORS requests from anywhere.

bear454:
As a developer of a web application that consumes _pint_ data, I require the API to implement Cross-Origin Resource Sharing (CORS).

As the _pint_ service is read-only, there's no risk in any web client consuming it (e.g. a client cannot be 'tricked' into pushing bad data, as nothing can be pushed), and credentials are not used, CORS can be enabled without limitation using a wildcard header: `Access-Control-Allow-Origin: *`, as defined in https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin .

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -183,6 +183,10 @@ class PublicCloudInfoSrv < Sinatra::Base
     end.doc.css(tag)
   end
 
+  before do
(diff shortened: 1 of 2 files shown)


bear454:
I'll try to get this deployed today.

Issue #9, opened 2020-09-09
no region information for frameworks where images are global


rjschwei:
When we publish image information for a framework that has global images and no infrastructure we cannot provide user data for the regions API. For example:

https://susepubliccloudinfo.suse.com/v1/oracle/regions.json

returns an empty. We either need to provide generic information, such as "No region information available. Images have the same identifier in all regions", or we need to have some way of adding the region information outside of images and server descriptions.

rjschwei:
This is partially related to #8

rjschwei:
AFter further consideration this is a client side issue see comments in #11

Issue #8, opened 2020-09-02
Handle google regions for image queries


rjschwei:
Now that the client and server can provide region information it is more obvious that the data for Google is inconsistent.

Because Google treats images by name and the name of images in all regions is the same the Google data has no region specific entries for images. However this results in empty queries when on the client side

```pint google images --active --region europe-north1```

is used. Now that the query of region names is possible it is obvious that this inconsistency exists. Server side we should handle this and return the proper image, i.e. for Google ignore the region setting and treat it as "global".

jeremy-moffitt:
I believe the PR from Charu https://github.com/SUSE-Enceladus/public-cloud-info-service/pull/10 resolves this issue. Closing the ticket.

PR #10, opened 2020-09-10
Handle google regions for image queries (CSE-65)

chajain:
when querying the public-cloud-info-service the region should be ignored when the CSP is Google

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -261,6 +261,11 @@ class PublicCloudInfoSrv < Sinatra::Base
         .in_state(params[:image_state])
         .in_env(environment)
         .set_region(params[:region])
+    elsif images(params[:provider])
+          .in_state(params[:image_state])
+          .in_region(params[:region]).empty?
+      images(params[:provider])
+        .in_state(params[:image_state])
     else
       images(params[:provider])
         .in_state(params[:image_state])
--- a/spec/features/app_spec.rb
+++ b/spec/features/app_spec.rb
@@ -613,3 +613,16 @@ describe 'metadata APIs' do
     end
   end
 end
+
+# we test the region images queries for google
+describe 'google region images query' do
+  describe 'google images' do
+    before do
+      @path = '/v1/google/europe-north1/images/active.xml'
+    end
+
+    it 'matches google images list' do
+      compare_with_fixture(@path)
+    end
+  end
+end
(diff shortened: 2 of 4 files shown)


spec/features/app_spec.rb
```diff
@@ -613,3 +613,16 @@
     end
   end
 end
+
+# we test the region images queries for google
+describe 'google region images query' do
+  describe 'google images' do
+    before do
+      @path = '/v1/google/europe-north1/images/active.xml'
```
guangyee, line 621: What happens if there are two regions with different active.xml? Or we don't have such scenario?

jeremy-moffitt: my understanding is that the Google images are global, so there is no region metadata for them, which causes the list to come back empty if a region is specified. See https://github.com/SUSE-Enceladus/public-cloud-info-service/issues/8  , so in this case if there are multiple regions, their active images should be the same.

guangyee (approved):
Sounds good.

app.rb
```diff
@@ -261,6 +261,9 @@ def form_xml_for_settings(tag, values)
         .in_state(params[:image_state])
         .in_env(environment)
         .set_region(params[:region])
+    elsif params[:provider] == 'google'
```
bear454, line 264: I don't like this being keyed on `'google'`; this should be a more generic behavior for any images that do not have a defined region or environment.

bear454 (changes requested):
Please make this more general for the case of non-regioned images instead of specific to provider==google.

app.rb
rjschwei: Yep, same applies to Oracle and this is related to #9 when we have global images and no update infrastructure.

spec/features/app_spec.rb
rjschwei: There is always only 1 active image for a given image type. For CSPs where images are global (Google and Oracle) the global image names that match the filters should be returned when a specific region is queried.

rjschwei (changes requested):

app.rb
chajain: I have updated the code so that no longer google is no longer hard coded

rjschwei (approved):

bear454 (approved):

jeremy-moffitt (approved):
LGTM, thanks!

PR #11, opened 2020-09-21
no region information for frameworks where images are global (CSE-66)

chajain:
When we publish image information for a framework that has global images and no infrastructure we cannot provide user data for the regions API. In this case the response indicates to the user that regions are not supported.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -216,9 +216,16 @@ class PublicCloudInfoSrv < Sinatra::Base
     validate_params_ext
     validate_params_provider
 
-    responses = form_xml_for_settings(
-      'region', settings.regions[params[:provider]]
-    )
+    responses = if settings.regions[params[:provider]].empty?
+      form_xml_for_settings(
+        'region', Array['No region information available. '\
+        'Images have the same identifier in all regions']
+      )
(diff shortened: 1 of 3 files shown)


guangyee (commented):
Is the "region" for display purposes only? Otherwise, I don't think hardcoding it to static string is a good idea. Clients will now need to make the distinction between this static string and a real region name.

bear454:
I agree; I think it makes much more sense as an API to simply return an empty set.

jeremy-moffitt:
My understanding is it returns empty as-is, assuming thats correct, we need clarity about the filed issue https://github.com/SUSE-Enceladus/public-cloud-info-service/issues/9  that Charu is trying to address here. @rjschwei can you please clarify the desired output? Thanks!

rjschwei:
The key is consistency per framework. If I can query the regions API and get an answer than I need to be able use any of the returned values to query images even if images are no region local. That's the part that was fixed with #10 . I guess #9 is really a client side problem. As an API I agree @guangyee and @bear454 it is fair to return an empty list. Sorry for sending you on a wild goose chase.

PR #13, opened 2021-01-29
Pint service provides an endpoint with the current data version 

chajain:
Pint REST API to be able to query for version information so as to know when there has been an update to the backend data.
Uses the bumpversion version data from pint-data/-/blob/master/.bumpversion.cfg as the version output for this purpose
Added the unit teststo validate version endpoint

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -178,6 +178,27 @@ class PublicCloudInfoSrv < Sinatra::Base
     return regions
   end
 
+  def data_version_info
+    data_version = []
+    path_parts = []
+    return unless ENV['FRAMEWORKS']
+
+    if File.dirname(ENV['FRAMEWORKS']).include? 'pint-data'
+      path_parts = File.dirname(ENV['FRAMEWORKS']).partition('pint-data')
+    end
+    return if path_parts.empty?
+
+    path = path_parts[0].concat(path_parts[1]).concat('/.bumpversion.cfg')
+    File.open(path) do |f|
+      f.any? do |version|
+        if version.include?('current_version =')
+          data_version << version.split('=')[1].strip
+        end
+      end
+    end
+    return data_version
+  end
+
   def environments?(provider)
     framework = settings.frameworks[provider]
     return !framework.css('environments').empty?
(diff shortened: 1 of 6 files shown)


jeremy-moffitt (commented):
With the huge caveat that I am not very familiar with the code base, I don't understand why there needs to be a new copy of frameworks.xml under fixtures/pint-data/data , guessing I'm overlooking something but would appreciate a brief explanation so I can learn something, thanks!

chajain:
> With the huge caveat that I am not very familiar with the code base, I don't understand why there needs to be a new copy of frameworks.xml under fixtures/pint-data/data , guessing I'm overlooking something but would appreciate a brief explanation so I can learn something, thanks!
spe
Since the current_version needs to be picked up from pint-data/.bumpversion.cfg for the new endpoint /v1/data-version.xml, the path is generated from ENV[FRAMEWORKS] which should have pint-data in the path. That is the reason I copied fixtures/framework.xml to fixtures/pint-data/data/framework.xml and updated ENV[FRAMEWORKS] in spec/spec_helper.rb .

app.rb
```diff
@@ -178,6 +178,27 @@ def get_alternate_region(provider, region)
...
+    data_version = []
+    path_parts = []
+    return unless ENV['FRAMEWORKS']
+
+    if File.dirname(ENV['FRAMEWORKS']).include? 'pint-data'
+      path_parts = File.dirname(ENV['FRAMEWORKS']).partition('pint-data')
+    end
+    return if path_parts.empty?
+
+    path = path_parts[0].concat(path_parts[1]).concat('/.bumpversion.cfg')
```
rjschwei, line 191: Nope, bumpversion.cfg is a file fo rthe bumpversion utilitiy it is strictly a development file and from another project. We need a different solution, we cannot depend on "internal" data.

I suggest to create a version.xml file or similar in the pint-data that then can be read by the server.

rjschwei (changes requested):

app.rb
bear454: bumpversion can update multiple destinations with the version, though, so it could write to XML as well as the specfile.

jeremy-moffitt:
with the change to pint-NG, is this PR still relevant? if not, we should close it

rjschwei:
Implementation moved to Python and we have a solution for the version query.

PR #14, opened 2021-03-02
uses github actions for CI rather than travis (CSE-172)

chajain:


.github/workflows/ruby.yml
```diff
@@ -0,0 +1,36 @@
...
+  test:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Ruby
+    # To automatically get bug fixes and new Ruby versions for ruby/setup-ruby,
+    # change this to (see https://github.com/ruby/setup-ruby#versioning):
+    # uses: ruby/setup-ruby@v1
```
jeremy-moffitt, line 27: is the specific Ruby version to avoid accidentally having it break due to breaking changes in Ruby itself? Is there anything particular about `473e4d8fe5dd94ee328fdfca9f8c9c7afc9dae5e` ?

chajain: When I set up continuous integration (CI) using GitHub Actions, it automatically creates a template for ruby and it automatically added that commit. It refers to this https://github.com/ruby/setup-ruby/releases/tag/v1.61.0. I can put v1 if it doesn't look right

guangyee: This is not the ruby version, rather, it is the setup-ruby action version. The actual ruby version will be taken from here https://github.com/SUSE-Enceladus/public-cloud-info-service/blob/master/.ruby-version according to the doc.

jeremy-moffitt: ah ok, thanks for clarifying... I dont know enough about it to say whether it looks right, don't block the change on this question for sure

bear454:
This should have triggered a github actions report... I'll look and see if we need to set something in the settings.

bear454:
https://github.com/SUSE-Enceladus/public-cloud-info-service/runs/2058445826?check_suite_focus=true

PR #20, opened 2021-04-01
Update get_provider_regions

mbelur:
Call to /v1/alibaba/regions call was failing since we do not have alibaba defined in PROVIDER_SERVERS_MODEL_MAP. 
We cannot rely only on the PROVIDER_SERVERS_MODEL to get the regions as in this case.
Code updated to use both PROVIDER_SERVERS_MODEL and PROVIDER_IMAGES_MODEL

### DIFF
--- a/next_generation/app.py
+++ b/next_generation/app.py
@@ -144,7 +144,7 @@ def get_formatted_dict(obj, extra_attrs=None, exclude_attrs=None):
     obj_dict = {}
     for attr in obj.__dict__.keys():
         if exclude_attrs and attr in exclude_attrs:
-            contineu
+            continue
         elif attr[0] == '_':
             continue
         else:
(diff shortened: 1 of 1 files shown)


jeremy-moffitt (commented):
Are the regions only available as attributes on the servers and images? That seems like it might be a scaling issue later, though thats how it was already being implemented.

I assume there was a use case where the image regions and server regions were mismatched that required this change?

*edit* was submitting this comment as the PR was updated to mention the Alibaba scenario, is the image region set not sufficient for other providers?

mbelur:
> Are the regions only available as attributes on the servers and images? That seems like it might be a scaling issue later, though thats how it was already being implemented.
> 
regions are not only available as attributes on the servers and images but also in the microsoft case, its available as environment attributes. This change does not fix the microsoft case. A separate change will be done to handle the microsoft case.

> I assume there was a use case where the image regions and server regions were mismatched that required this change?
There were a couple of usecase, like the alibaba and oracle that were failing earlier

> _edit_ was submitting this comment as the PR was updated to mention the Alibaba scenario, is the image region set not sufficient for other providers?
We cannot only rely on the Image attribute also because in the case of region is only an attribute of the servers.

guangyee:
The Microsoft environment/region mapping schema still a TBD at this point. @rjschwei is working with @bear454 to finalize the design.

next_generation/app.py
```diff
@@ -362,7 +375,7 @@ def list_provider_resource_for_category(provider, region, category):
     assert_valid_category(category)
     resources = globals()['get_provider_%s_for_region' % (category)](
         provider, region)
-    return make_response(resources, category, catetory[:-1])
+    return make_response(resources, category, category[:-1])
```
guangyee, line 378: good one! Unit test rules! :-)

guangyee (approved):

PR #24, opened 2021-04-05
Add functional tests

mbelur:


next_generation/README.rst
```diff
@@ -32,9 +32,44 @@ Follow the following steps to run these unittests:
...
 
+
+Running the Functional Tests
+------------------------------
+Follow the following steps to run the functional tests:
+
+Pre-requisite
+These functional tests expect the environment under test to be setup correctly.
+
+1. Setup a python virtual environment
```
jeremy-moffitt, line 49: this currently has to be run from the next_generation folder, which presumably will be the top level folder eventually... though I suppose since this readme is in that folder, I probably should have assumed as much

jeremy-moffitt (approved):
worked for me, I only ran with the default URL though

chajain (approved):
Looks Good

jeremy-moffitt:
tested again with the pint-ng server running locally, got the expected results :)

next_generation/README.rst
guangyee: Yes, we'll eventually move it to the top level as the Sinatra code will be removed in favor of the Flask app.

guangyee (approved):
Look good

PR #28, opened 2021-04-15
use pint/pint-server-lambda-function from IBS

guangyee:
Use the SLE15-base AWS Lambda base continer image instead of the one
from AWS.

### DIFF
--- a/next_generation/Dockerfile.sle15
+++ b/next_generation/Dockerfile.sle15
@@ -1,0 +1,11 @@
+FROM registry.suse.de/home/yeey/pintserverng/containers/pint/pint-server-lambda-function:latest
+
+# To test development changes, you can just directly COPY files and layer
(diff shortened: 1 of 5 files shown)


next_generation/Dockerfile.sle15
```diff
@@ -0,0 +1,11 @@
+FROM registry.suse.de/home/yeey/pintserverng/containers/pint/pint-server-lambda-function:latest
```
jeremy-moffitt, line 1: This is the internal repo? It may be an issue referencing a private registry from a public repository. I dont have an alternative suggestion though

guangyee: I don't think we need to reference it from ECR. We'll upload the image directly to ECR for production deployment. Dockerfile.sle15 is used for internal development only, for instantly test code changes before updating the package in IBS.

PR #33, opened 2021-04-20
Update Makefile to check for system libraries

mbelur:
Also update Dockerfile to copy from the .staging
folder.

### DIFF
--- a/next_generation/Makefile
+++ b/next_generation/Makefile
@@ -14,10 +14,17 @@ build_psycopg2_bin:
 	python -m pip install psycopg2 -t ".staging/opt/python/lib/python3.6/site-packages"
 
 copy_dependent_libs:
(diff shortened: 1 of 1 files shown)


next_generation/Dockerfile.aws
```diff
@@ -6,8 +6,8 @@ RUN mkdir -p ./pint_server
 COPY pint_server/. ./pint_server
 
 
-COPY opt/. /opt
-COPY usr/. /usr
+COPY .staging/opt/. /opt
+COPY .staging/usr/. /usr
```
guangyee, line 10: these two are not needed as the Dockerfile is used inside the .staging dir

mbelur: Updated

guangyee (approved):
looks good

PR #34, opened 2021-04-20
improve version and provider lookup

guangyee:
NOTE: this PR must be merged in conjunction with
https://gitlab.suse.de/public-cloud-team/pint-data/-/merge_requests/68

### DIFF
--- a/next_generation/pint_server/app.py
+++ b/next_generation/pint_server/app.py
@@ -1,4 +1,5 @@
 import datetime
+import re
 from decimal import Decimal
 from flask import abort, Flask, jsonify, make_response, request, redirect, \
                   Response
@@ -6,6 +7,7 @@ from flask_cors import CORS
 from flask_sqlalchemy import SQLAlchemy
 from sqlalchemy import text, or_
 from sqlalchemy.exc import DataError
+from sqlalchemy.orm.exc import NoResultFound, MultipleResultsFound
 from xml.dom import minidom
 import xml.etree.ElementTree as ET
 
@@ -57,33 +59,16 @@ def get_supported_providers():
     global CACHE_PROVIDERS
 
     if CACHE_PROVIDERS is None:
-        CACHE_PROVIDERS = query_providers()
+        versions  = VersionsModel.query.with_entities(VersionsModel.tablename)
+        CACHE_PROVIDERS = list(
+            {re.sub('(servers|images)', '', v.tablename) for v in versions})
 
-    return CACHE_PROVIDERS
-
-
-def query_providers():
-    """Get all the providers"""
-    global CACHE_PROVIDERS
-
-    if CACHE_PROVIDERS is not None:
-        return CACHE_PROVIDERS
-
-    # FIXME(gyee): this query is very specific to PostgreSQL. If we support
-    # other DBs such as MySQL, we'll need to conditionally change this.
-    # Ideally, this information should be in a SQL lookup table so it's
-    # database type agnostic.
-    sql_stat = text("select regexp_replace(table_name, 'servers|images', '') "
-                    "from information_schema.tables where "
-                    "table_schema = 'public' and table_name like '%images'")
-    result = db_session.get_bind().execute(sql_stat)
-    CACHE_PROVIDERS = [row[0] for row in result]
     return CACHE_PROVIDERS
 
 
 def get_providers():
-    providers = query_providers()
-    return [{'name': provider} for provider in providers]
+    """Get all the providers"""
+    return [{'name': provider} for provider in get_supported_providers()]
 
 
 def json_to_xml(json_obj, collection_name, element_name):
@@ -323,13 +308,16 @@ def get_provider_images(provider):
 
 
 def get_data_version_for_provider_category(provider, category):
-    column_name = provider + category
-    versions = VersionsModel.query.all()[0]
+    tablename = provider + category
     try:
-        getattr(versions, column_name)
-    except AttributeError:
+        version = VersionsModel.query.filter(
+            VersionsModel.tablename == tablename).one()
+    except (NoResultFound, MultipleResultsFound):
+        # NOTE(gyee): we should never run into MultipleResultsFound exception
+        # or otherse we have data corruption problem in the database.
         abort(Response('', status=404))
-    return {'version': str(float(getattr(versions, column_name)))}
+
+    return {'version': str(version.version)}
 
 
 def assert_valid_provider(provider):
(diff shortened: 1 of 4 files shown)


next_generation/pint_server/models.py
```diff
@@ -108,11 +108,5 @@ class MicrosoftRegionMapModel(Base, PintBase):
...
-    amazonservers = Column(Numeric, nullable=False, primary_key=True)
-    amazonimages = Column(Numeric, nullable=False)
-    googleservers = Column(Numeric, nullable=False)
-    googleimages = Column(Numeric, nullable=False)
-    oracleimages = Column(Numeric, nullable=False)
-    microsoftservers = Column(Numeric, nullable=False)
-    microsoftimages = Column(Numeric, nullable=False)
-    alibabaimages = Column(Numeric, nullable=False)
+    tablename = Column(String(50), primary_key=True)
+    version = Column(Numeric)
```
rtamalin, line 112:
```suggestion
    version = Column(Numeric, nullable=False)
```

next_generation/pint_server/app.py
```diff
@@ -323,13 +311,13 @@ def get_provider_images(provider):
...
 
 def get_data_version_for_provider_category(provider, category):
-    column_name = provider + category
-    versions = VersionsModel.query.all()[0]
-    try:
-        getattr(versions, column_name)
-    except AttributeError:
+    tablename = provider + category
+    version = VersionsModel.query.filter(
+        VersionsModel.tablename == tablename).first()
```
rtamalin, line 316: Maybe consider using `one()` instead of first here which raises exceptions if there are None or more than one... Could allow you to simplify this routine to be just a return statement with exception handling that calls abort for the None, or multiples found cases... Just a thought ;-)

next_generation/pint_server/app.py
```diff
@@ -57,33 +58,20 @@ def get_supported_providers():
     global CACHE_PROVIDERS
 
     if CACHE_PROVIDERS is None:
-        CACHE_PROVIDERS = query_providers()
+        versions  = VersionsModel.query.with_entities(VersionsModel.tablename)
+        providers = []
```
rtamalin, line 62: This could probably be done with a simple set comprehension, converted to a list at the end if you need CACHE_PROVIDERS to be a list, rather than a set... E.g.

 
```suggestion
        providers = list({re.sub('(servers|images)', '', v.tablename)
                          for v in versions})
```

rtamalin (changes requested):
Looks good; made a couple of suggestions that can be ignored.

However the nullable=False suggestion should be applied to ensure that the version value can never be a null value in the versions table.

next_generation/pint_server/models.py
```diff
@@ -108,11 +108,5 @@ class MicrosoftRegionMapModel(Base, PintBase):
...
 
-    amazonservers = Column(Numeric, nullable=False, primary_key=True)
-    amazonimages = Column(Numeric, nullable=False)
-    googleservers = Column(Numeric, nullable=False)
-    googleimages = Column(Numeric, nullable=False)
-    oracleimages = Column(Numeric, nullable=False)
-    microsoftservers = Column(Numeric, nullable=False)
-    microsoftimages = Column(Numeric, nullable=False)
-    alibabaimages = Column(Numeric, nullable=False)
+    tablename = Column(String(50), primary_key=True)
```
rtamalin, line 111: Here you have the length as 50 but in https://gitlab.suse.de/public-cloud-team/pint-data/-/merge_requests/68 you have it as 100...

```suggestion
    tablename = Column(String(100), primary_key=True)
```

next_generation/pint_server/models.py
rjschwei: Yep, there must be a version, cannot be null.

next_generation/pint_server/app.py
```diff
@@ -57,33 +58,20 @@ def get_supported_providers():
...
+        for version in versions:
+            provider = re.sub('(servers|images)', '', version.tablename)
+            if provider not in providers:
+                providers.append(provider)
+        CACHE_PROVIDERS = providers
```
rjschwei, line 67: Well caching is somewhat of an odd concept to use in the way we are deploying the code. Every query will trigger the lambda function and each time it runs it starts from scratch, i.e. it is a stateless implementation. As such a cache is only usefull if in the same execution of a query the providers list needs to be accessed multiple times.

next_generation/pint_server/models.py
guangyee: good catch!

next_generation/pint_server/app.py
guangyee: The reason I used first() here is because I want to explicitly check for no result so we can return a 404 to match the current Pint Server behavior.

next_generation/pint_server/app.py
guangyee: We want the values to be unique. Otherwise, we'll end up with two 'amazon' entries for 'amazonservers' and 'amazonimages'.

next_generation/pint_server/models.py
guangyee: good catch!

next_generation/pint_server/app.py
guangyee: When I first implemented this, I was thinking the possibility of running this code outside of AWS Lambda. James mentioned that we want this to be portable not just between cloud providers, but we want it to be able to independently hosted by an open source project. This Flask app can also be running with Apache2 WSGI.

next_generation/pint_server/app.py
rtamalin: My suggestion used a "set" comprehension, which means that there can be no duplicates in the resulting list, since the set would eliminate them ;-)

next_generation/pint_server/app.py
rtamalin: That is the point of the exception handing of `one()` - https://docs.sqlalchemy.org/en/12/orm/query.html#sqlalchemy.orm.query.Query.one

It raises sqlalchemy.orm.exc.NoResultFound for no results found, and sqlalchemy.orm.exc.MultipleResultsFound for multiple results found, so you can handle both scenarios appropriately using exception handling without need to do an explicit check.

rtamalin (commented):
See my follow on comments for the two unresolved threads... I don't think grok'd what I was getting at the first time around.

next_generation/pint_server/app.py
guangyee: Sound good. Let me give it a try.

next_generation/pint_server/app.py
guangyee: Let me give it a try.

guangyee: works! Thanks!

next_generation/pint_server/models.py
rtamalin: Looks good!

rtamalin (approved):

rjschwei (approved):

PR #37, opened 2021-04-29
load generator script

mbelur:
This script hits the pint service API
endpoints with user entered number of requests
spread over the user supplied duration


jeremy-moffitt (changes requested):
One change request for this version and a couple of suggestions for future enhancements...
1. For the base url , please add the option to override it with a parameter... so it should default to localhost:5000 but we should be able to point it at a different url via command-line parameter rather than needing to edit the python file

Future enhancements:
1. check the response time of each request, and share out a summary of the data after the test completes (i.e. "100 requests completed in < 0.5 seconds, 500 requests were >=0.5 & <1.0sec, 100 requests were 1.0-1.99 seconds, 3 requests > 2.0 seconds)
2. randomize the provider in the images and servers list, so its not always asking for the Amazon data, in the production environment caching will come into play, and mixing it up a bit will help exercise that

next_generation/pint_server/tests/loadtest/loadtest.py
```diff
@@ -0,0 +1,62 @@
...
+import random
+import requests
+import sys
+import threading
+import time
+
+'''
+To run this program, example: python loadtest.py --num_requests 100 --duration 60
+'''
+base_url = "http://localhost:5000"
```
guangyee, line 13: I agree the base_url should be configurable.

next_generation/pint_server/tests/loadtest/loadtest.py
```diff
@@ -0,0 +1,62 @@
...
+urls_to_hit = ["/v1/providers", "/v1/amazon/images", "/v1/amazon/servers"]
```
guangyee, line 14: maybe add more APIs?

next_generation/pint_server/tests/loadtest/loadtest.py
```diff
@@ -0,0 +1,62 @@
...
+from datetime import datetime, timedelta
+import getopt
+import random
+import requests
+import sys
+import threading
+import time
+
+'''
+To run this program, example: python loadtest.py --num_requests 100 --duration 60
```
guangyee, line 11: duration in minutes right?

next_generation/pint_server/tests/loadtest/loadtest.py
mbelur: sure will update it

next_generation/pint_server/tests/loadtest/loadtest.py
mbelur: will update to add more providers in the mix

next_generation/pint_server/tests/loadtest/loadtest.py
mbelur: this is in seconds..

mbelur:
> One change request for this version and a couple of suggestions for future enhancements...
> 
> 1. For the base url , please add the option to override it with a parameter... so it should default to localhost:5000 but we should be able to point it at a different url via command-line parameter rather than needing to edit the python file
> 
> Future enhancements:
> 
> 1. check the response time of each request, and share out a summary of the data after the test completes (i.e. "100 requests completed in < 0.5 seconds, 500 requests were >=0.5 & <1.0sec, 100 requests were 1.0-1.99 seconds, 3 requests > 2.0 seconds)
> 2. randomize the provider in the images and servers list, so its not always asking for the Amazon data, in the production environment caching will come into play, and mixing it up a bit will help exercise that

Sure will make the base url a command line option and use localhost as the default if none provided.
You can use the time command to get the data when you kick off the script.
Yeah will add additional providers to the mix for images and servers API call.

jeremy-moffitt:
> You can use the time command to get the data when you kick off the script.

I'm a bit confused on this one, the time command will measure each threads total time? I thought it would only measure the entire script start to finish

mbelur:
> > You can use the time command to get the data when you kick off the script.
> 
> I'm a bit confused on this one, the time command will measure each threads total time? I thought it would only measure the entire script start to finish

time command for the entire script execution.

jeremy-moffitt:
> time command for the entire script execution. it might not completely match with the duration you provided but will be close.

Thats different than what I was thinking... Note that I don't think we need this for the first version of this, but for clarity sake will describe what I was thinking below (mix of actual and pseudocode):
```
///somewhere global - create a dictionary of response time ranges
responseTimes = {}; 

///inside the task - grab the response time off of each request as it finishes
response = requests.get(url, verify=False)
/// bucket by 0.5 second increments, by div 0.5 and rounding down
timeBucketIndex = response.elapsed // 0.5
/// if a bucket isnt initialized yet, set it to 1 since a request came in in that bucket
if responseTimes[timeBucketIndex] is None: 
  responseTimes[timeBucketIndex] = 1
else:
/// if the bucket already has a value, increment it
  responseTimes[timeBucketIndex] += 1

// somewhere global after all threads are done - print out the list of response times and how many requests fell into each bucket
for time,count in responseTimes.items():
  print((time * 0.5, " - " , (time + 1) * 0.5 , " : ", count)

```

mbelur:
> > time command for the entire script execution. it might not completely match with the duration you provided but will be close.
> 
> Thats different than what I was thinking... Note that I don't think we need this for the first version of this, but for clarity sake will describe what I was thinking below (mix of actual and pseudocode):
> 
> ```
> ///somewhere global - create a dictionary of response time ranges
> responseTimes = {}; 
> 
> ///inside the task - grab the response time off of each request as it finishes
> response = requests.get(url, verify=False)
> /// bucket by 0.5 second increments, by div 0.5 and rounding down
> timeBucketIndex = response.elapsed // 0.5
> /// if a bucket isnt initialized yet, set it to 1 since a request came in in that bucket
> if responseTimes[timeBucketIndex] is None: 
>   responseTimes[timeBucketIndex] = 1
> else:
> /// if the bucket already has a value, increment it
>   responseTimes[timeBucketIndex] += 1
> 
> // somewhere global after all threads are done - print out the list of response times and how many requests fell into each bucket
> for time,count in responseTimes.items():
>   print((time * 0.5, " - " , (time + 1) * 0.5 , " : ", count)
> ```

Yeah we can do these as enhancements in the next iteration

jeremy-moffitt:
I do think you need to rebase this now that the cutover has merged, once thats done it looks good to me

mbelur:
> I do think you need to rebase this now that the cutover has merged, once thats done it looks good to me

done

jeremy-moffitt (approved):
LGTM!

pint_server/tests/loadtest/loadtest.py
```diff
@@ -0,0 +1,85 @@
...
+                        type=int,
+                        default=100,
+                        dest="num_requests",
+                        help="Number of requests to hit the API")
+    parser.add_argument('-d', '--duration',
+                        required=False,
+                        type=int,
+                        default=60,
+                        dest="duration",
+                        help="Duration in seconds")
```
guangyee, line 47: I think minutes would be better as load tests usually run in minutes.

mbelur: Updated the script to use minutes for duration

guangyee (approved):
nice work!

PR #38, opened 2021-04-29
update README.rst and SUSE.rst

guangyee:
1. correct a typo in README.rst
2. update SUSE.rst with correct instrustions

### DIFF
--- a/Dockerfile.sle15
+++ b/Dockerfile.sle15
@@ -1,4 +1,4 @@
-FROM registry.suse.de/home/yeey/pintserverng/containers/pint/pint-server-lambda-function:latest
+FROM registry.suse.de/devel/pubcloud/productionservices/pint/containers/pint/pint-server-lambda-function:latest
 
 # To test development changes, you can just directly COPY files and layer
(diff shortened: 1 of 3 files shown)


Dockerfile.sle15
```diff
@@ -1,4 +1,4 @@
-FROM registry.suse.de/home/yeey/pintserverng/containers/pint/pint-server-lambda-function:latest
+FROM registry.suse.de/devel/pubcloud/productionservices/pint/containers/pint/pint-server-lambda-function:latest
```
jeremy-moffitt, line 1: this was confusing to me at first since you cant navigate to this URL directly, but it does show up with this path at https://registry.suse.de/cgi-bin/cooverview  ... its a shame the search function on that page doesnt work right

jeremy-moffitt (approved):
LGTM

Dockerfile.sle15
rjschwei: @guangyee missed that. This shouldn't even be here, it reveals SUSE internal information in that it shows that we run an internal registry on that server. We need to come up with a different approach.

PR #39, opened 2021-04-29
Summarize response times at end of load test

jeremy-moffitt:
Track the response times of the requests from the load generator and print a summary at the end

### DIFF
--- a/pint_server/tests/loadtest/loadtest.py
+++ b/pint_server/tests/loadtest/loadtest.py
@@ -25,11 +25,28 @@ urls_to_hit = ["/v1/providers",
                "/v1/providers.xml",
                "/v1/alibaba/images.xml"]
 
+time_buckets = {}
+
 def task(base_url):
     print("Executing the Task")
     url = base_url + random.choice(urls_to_hit)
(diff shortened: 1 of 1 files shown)


guangyee (approved):
looks good

mbelur (commented):
Looks Good. Should we keep track of the response time based on the API call type separately to give us average response time data specific to the type of call?

jeremy-moffitt:
> Looks Good. Should we keep track of the response time based on the API call type separately to give us average response time data specific to the type of call?

That sounds like a reasonable next enhancement. If this looks ok as-is, maybe we merge this one and then as time allows I can add that improvement (also pondering adding a little ascii graph, but that might be beyond my abilities ;-))

mbelur:
> > Looks Good. Should we keep track of the response time based on the API call type separately to give us average response time data specific to the type of call?
> 
> That sounds like a reasonable next enhancement. If this looks ok as-is, maybe we merge this one and then as time allows I can add that improvement (also pondering adding a little ascii graph, but that might be beyond my abilities ;-))

sounds good.

mbelur (approved):

PR #41, opened 2021-05-18
Add orm_exporttopostgres to bin/

rtamalin:
Port the orm_exporttopostgres script to work with the new pint_server
implementation.

To use it you should clone the pint-data repo, that contains the XML
data to be import, beside the public-cloud-info-service; the script
will automatically look for the pint-data repo relative to it's own
location in the public-cloud-info-service/bin directory.

Run it by specifying the necessary command line arguments, e.g.

    % bin/orm_exporttopostgres --host ${POSTGRES_HOST} \
        --password ${POSTGRES_PASSWORD}
        --user ${POSTGRES_USER}

The script will create the database tables if not already created,
and will then import the XML content from the pint-data repo, either
as a full import if no data exists, or incrementally importing new
changes made since the last import.

For incremental updates, when updating the versions table, if the
existing version entry value is less than the detected commit date
stamp for the updates, then we will use the new commit date stamp,
otherwise we will add 0.01 to the existing date stamp value. This
should allow us to support up to 100 incremental merges for the
same date without issues, and will still work if we ever have more
than that number of incremental updates, though the relationship
between the version value and the associated commit data may be
lost.

To test the incremental update support I used the following shell
command when cd'd into a pint data repo clone, with this repo
cloned beside it:

    % for i in $(seq 90 -1 0); do git co origin/master~${i} && \
      ../public-cloud-info-service/bin/orm_exporttopostgres \
          --host ${POSTGRES_HOST} \
          --password ${POSTGRES_PASSWORD} \
          --user ${POSTGRES_USER}; done

Include lmxl (v4.6.3) in the requirements.txt to support running
the orm_exporttopostgres script.

### DIFF
--- a/bin/orm_exporttopostgres
+++ b/bin/orm_exporttopostgres
@@ -1,0 +1,361 @@
+#!/usr/bin/env python3
+
+# Script to import the XML based pint data into a PostgreSQL DB
+# THis is a one time operation, the database is expceted to hav a table called
+# "postgres" that has no entries
+
+import argparse
+from datetime import datetime
+import glob
+import os
+import subprocess
+import sys
+
+from lxml import etree
+
+from pint_server.database import init_db
+from pint_server.models import (
+            ImageState,
+            ServerType,
+            AlibabaImagesModel,
+            AmazonImagesModel,
+            AmazonServersModel,
+            GoogleImagesModel,
+            GoogleServersModel,
+            MicrosoftImagesModel,
+            MicrosoftRegionMapModel,
+            MicrosoftServersModel,
+            OracleImagesModel,
+            VersionsModel
+        )
+
+import logging
+
+
+def gen_data_files_list(pint_data_repo=None):
+    
+    if pint_data_repo is None:
+        pint_data_repo = os.path.realpath(
+                             os.path.join(
+                                 os.path.dirname(os.path.abspath(__file__)),
+                                 "..", "..", "pint-data"))
+
+    logger.debug("Pint Data repo: %s", repr(pint_data_repo))
+
+    data_files_pattern = os.path.join(pint_data_repo, 'data', '*.xml')
+    logger.debug("Data Files pattern: %s", repr(data_files_pattern))
+
+    data_files = glob.glob(data_files_pattern)
+    logger.debug("Data Files: %s", repr(data_files))
+
+    return data_files
+
+def get_commit_date(data_file):
+    data_dir = os.path.dirname(data_file)
+
+    output = subprocess.run(["bash", "-c",
+                             "cd %s; git log -n1 --pretty='format:%s' "
+                             "--date='format:%s' %s" %
+                             (os.path.dirname(data_file), '%cd', '%Y%m%d',
+                              os.path.basename(data_file))],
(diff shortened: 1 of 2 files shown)


bin/orm_exporttopostgres
```diff
@@ -0,0 +1,341 @@
...
+
+    data_files_pattern = os.path.join(pint_data_repo, 'data', '*.xml')
+    logger.debug("Data Files pattern: %s", repr(data_files_pattern))
+
+    data_files = glob.glob(data_files_pattern)
+    logger.debug("Data Files: %s", repr(data_files))
+
+    return data_files
+
+def get_commit_date(data_file):
```
guangyee, line 53: What if we have multiple commits in a day? According to Robert, that's a possibility.

bin/orm_exporttopostgres
```diff
@@ -0,0 +1,341 @@
...
+        #con.commit()
+    return rows
+
+def extract_data_from_file(data_file, data_store):
+
+    commit_date = get_commit_date(data_file)
+    reg_map = None
+
+    provider = os.path.basename(data_file).split('.')[0]
+    data_store[provider] = dict(version=f"{commit_date}.0",
```
guangyee, line 114: If we have multiple commits a day, which is possible, we should bump the last digit instead of always hardcoding it to '.0'.

guangyee (changes requested):
Per my understanding, there's a possibility of multiple commits in a day. If that's the case, we need to increment the last digit in the version.

bin/orm_exporttopostgres
rtamalin: So using the commit time stamp info isn't perfect in and of itself anyway, as you can merge commits out of order, so a change that was proposed yesterday could be merged after one that was proposed this morning, and you end up with the version date being yesterday.

I used the commit date instead of just the runtime date stamp to allow me to track the relationship between the incremental merges as I made them in testing...

However we could switch back to the original model of using the timestamp from when the command is run.

rtamalin: I believe I have addressed this issue in my latest change, as well as removing some redundant code, and fixing a bug related to not using the correct table name when doing the existing version query.

Basically we will take the "suggested" version, and if it is newer than the existing entry, we will use it, otherwise we will increment the existing value by 0.01, which will handle older commit date stamps being used due to out of order merges as well as up to 100 incremental updates per day while still maintaining the logical relationship between commit date stamp and version, and will still support >100 incremental updates per-day with that logical relationship being lost due to the version value "rolling over" to the next logical day...

bin/orm_exporttopostgres
rtamalin: As per the discussion above I believe this is also addressed in an appropriate fashion, with the comments around this code being updated to reflect that this version is a "suggested" one, rather than the version that will actually be set.

rtamalin:
Updated description to reflect the fact that I ran the tests for the last 91 (rather than 3) commits; cut and paste wrong command line history entry ;-)

guangyee (approved):
LGTM

mbelur (approved):
LGTM!

guangyee (approved):

PR #43, opened 2021-05-28
account for Azure environment with query images for region

guangyee:


mbelur (approved):
LGTM!

jeremy-moffitt (approved):
LGTM!

mbelur:
Actually there is a count mismatch between the existing pint and the pint-ng for microsoft regions:

Actually the count of regions do not seem to match:

curl https://susepubliccloudinfo.suse.com/v1/microsoft/regions | python -mjson.tool | grep 'name' | wc -l
156

curl http://localhost:5000/v1/microsoft/regions | python -mjson.tool | grep 'name' | wc -l
67

mbelur (approved):
LGTM! We are looking at a data issue with one of the microsoft regions - southcentralusstg. 
However, this change looks good to me. we can merge this change.

jeremy-moffitt (approved):

PR #47, opened 2021-06-03
fix data discrepancy in the output with google/<region>/images/<imgst…

mbelur:
…ate> call

Google images do not have region but instead project.
However, while filtering the images with a certain image state on a specific region,
the existing pint did return images of the image state. This change
does the same for the pint-ng


guangyee (approved):
thanks

PR #48, opened 2021-06-04
fix discrepancies between Pint and Pint NG

guangyee:
1. if 'changeinfo' is empty, don't return it
2. if 'urn' is empty, don't return it
3. for Alibaba and Oracle where we don't have any servers, we'll return
   an empty list when query by server type in order to be backward
   compatible
4. add mapping to map the original server types (i.e. smt, regionserver-sles,
   regionserver-sap, etc) to the new server types (i.e. update and region)
   in order to maintain backward compatibility


mbelur (approved):
LGTM!

PR #49, opened 2021-06-05
using the server's 'shape' attribute to facilitate backward compatibi…

guangyee:
…lity

The API's will return <type>-<shape> if <shape> is not empty. For example,
if the given server has type 'update' and shape 'sap', the API will return
'smt-sap' in it's server type. If <shape> is empty, 'smt' will be returned
instead.


rjschwei (approved):
LGTM

rtamalin (approved):
LGTM - thanks @guangyee

PR #51, opened 2021-06-08
add schema versioning and migration framework

guangyee:
The schema versioning and migration is facilitated by sqlalchemy-migrate[1].

Also add a CLI (implemented using python-click) to simpify schema and data
upgrade. With the CLI, we no longer need orm_exporttogres script and the code
have be incorporated into the CLI as well. To perform schema and data upgrade,
we simply need to invoke the pint_db_migrate.sh CLI. For example:

./bin/pint_db_migrate.sh -h db_host -U db_user -W db_password upgrade

TODO: We will need to remove the orm_exporttopostgres script once we are OK
with this PR.

1. https://sqlalchemy-migrate.readthedocs.io/en/latest/api.html


pint_server/pint_db_migrate/versions/001_in_the_beginning.py
```diff
@@ -0,0 +1,153 @@
...
+from sqlalchemy.dialects import postgresql
+
+
+meta = MetaData()
+
+server_type = postgresql.ENUM('region', 'update', name='server_type',
+                               metadata=meta)
+
+image_state = postgresql.ENUM('deleted', 'deprecated', 'inactive',
+                             'active', name='image_state', metadata=meta)
```
rtamalin, line 29: Missing one character of indent to align start of `'active'` under start of `'deleted'`

pint_server/pint_db_migrate/README
```diff
@@ -0,0 +1,4 @@
+This is a database migration repository.
+
+More information at
+http://code.google.com/p/sqlalchemy-migrate/
```
rtamalin, line 4: Probably should also add some details about how this is expected to be used for the pint DB project specific, e.g. a high level overview of the workflow for those not familiar with it ;-)

rtamalin (changes requested):

rtamalin:
Sorry @guangyee I managed to click the request changes before entering my final comments...

So most of the things in my require are just minor nits or only suggestions for better shell scripting.

There is one whitespace issue that should be addressed, and the README.md should definitely be fleshed out with more details ;-)

bin/pint_db_migrate.sh
```diff
@@ -0,0 +1,34 @@
...
+#
+# To contact SUSE about this file by physical or electronic mail,
+# you may find current contact information at www.suse.com
+
+
+MYNAME=$0
+CURRENTDIR=$(dirname $(readlink -e $MYNAME))
+REPOSITORY_DIR=$(dirname $CURRENTDIR)/pint_server/pint_db_migrate
+
+export PYTHONPATH=$(dirname $CURRENTDIR):$PYTHONPATH
```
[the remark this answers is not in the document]
guangyee: good point!

bin/pint_db_migrate.sh
```diff
@@ -0,0 +1,34 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE about this file by physical or electronic mail,
+# you may find current contact information at www.suse.com
+
+
+MYNAME=$0
```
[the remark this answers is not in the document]
guangyee: sounds good

pint_server/pint_db_migrate/manage.py
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env python
+from migrate.versioning.shell import main
+
+if __name__ == '__main__':
+    main(debug='False')
```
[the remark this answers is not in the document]
guangyee: this code is generated by the framework. Let me just remove it.

pint_server/pint_db_migrate/versions/001_in_the_beginning.py
guangyee: good catch!

pint_server/pint_db_migrate/README
guangyee: this file is generated by the framework. Let up update the README.rst at the top with the appropriate instructions instead.

bin/pint_db.py
```diff
@@ -0,0 +1,402 @@
+# Copyright (c) 2021 SUSE LLC
```
rtamalin, line 1: The following is not something I'm suggesting needs to be changed, just some feedback on the toolchain/workflow...


This script doesn't have a `#!` line, so it isn't really an "executable" so technically shouldn't be located in a `bin` directory... The original orm_exporttopostgres was an executable script because it was located in a `bin` directory...

Also everything that you are doing below in the wrapper shell script can be done in this python script itself, e.g. checking for the presence of the VIRTUAL_ENV env var, and dynamically updating the sys.path list to include the public-cloud-info-service...

And the latter would not be required if the create_dev_venv.sh script was modified to do a `pip install $SRCDIR` to ensure that the `pint_server` module is also installed into the virtualenv... Or better yet `pip install -e $SRCDIR`, i.e. developer mode install using a symlink to the $SRCDIR rathern than copying the contends, so that you don't have to keep re-installing if you are making development changes ;-)

In the long run it might be good to just make this script an executable script, which works with whatever the active runtime environment is. You could have the default value for the --repository option be the path that is dynamically calculated relative to the path of this script, similar to the determination of the `pint-data` repo location. ;-)

bin/pint_db_migrate.sh
```diff
@@ -0,0 +1,33 @@
...
+
+
+CurrentDir=$(dirname $(readlink -e $0))
+RepositoryDir=$(dirname $CurrentDir)/pint_server/pint_db_migrate
+
+export PYTHONPATH=$(dirname $CurrentDir):${PYTHONPATH:+:${PYTHONPATH}}
+
+if [[ "$VIRTUAL_ENV" == "" ]] ; then
+  echo
+  echo "ERROR: python virtualenv not detected. Please run $CurrentDir/create_venv.sh to create and activate the python venv first."
```
rtamalin, line 27: There isn't a `create_venv.sh` script (anymore?) but instead there is a `create_dev_venv.sh` and a `create_test_venv.sh` script. Guessing you mean the former?

requirements.txt
```diff
@@ -1,19 +1,24 @@
...
 Flask==1.1.2
-Flask-CORS==3.0.10
+Flask-Cors==3.0.10
 Flask-SQLAlchemy==2.4.4
 itsdangerous==1.1.0
 Jinja2==2.11.3
 jmespath==0.10.0
 lxml==4.6.3
 MarkupSafe==1.1.1
+pbr==5.6.0
```
rtamalin, line 13: Do we absolutely need pbr? ;-) Bad memories of debugging issues with it...

README.rst
```diff
@@ -6,6 +6,44 @@ Public Cloud Information Service enables users to lookup Public Cloud
 images and services information via REST API. Image and Server information
 is tracked in a PostgreSQL database.
 
+=============
```
rtamalin, line 9: Would also be a good idea to outline the workflow for changing the schema, even if just at a high level.

bin/pint_db_migrate.sh
guangyee: Good catch! I meant create_dev_env.sh. Let me update.

requirements.txt
guangyee: yes, for packaging.

README.rst
guangyee: sure, I'll add some notes on that.

bin/pint_db.py
guangyee: that's a good point. Let me see what I can do.

rtamalin (approved):
LGTM

PR #52, opened 2021-06-08
Implement init_db() enhancements in pint_server

rtamalin:
Add support to the pint_server init_db() function to take optional
dbconfig, outputfile and echo parameters.

The dbconfig paramater allows a dictionary to be specified that
holds the required config settings needed to contruct the connection
URL string. If a dbconfig is specified then it's settings will be
used in preference to those specified in the environment. Otherwise
the DATABASE_URI env var setting will be used, or if that is not
specified then the individual POSTGRES_... env var settings will be
used.

If the outputfile parameter is specified it will be used to create
a DB specific logging service to log output from SQLAlchemy DB ops.

The echo parameter will be passed as the echo parameter when calling
create_engine() allowing control over the associated engine logging
verbosity.

Updated pint_server/__init__.py to explicitly import the objects
that we want to expose from the pint_server module.

Updated PintBase to provide a tablename property.

### DIFF
--- a/pint_server/database.py
+++ b/pint_server/database.py
@@ -39,21 +39,59 @@ from pint_server.models import (
         VersionsModel
     )
 
-
 def get_environ_or_bust(key_name):
     assert key_name in os.environ, 'Environment variable %s is required.' % (
         key_name)
     return os.environ.get(key_name)
 
 
-def create_postgres_url_from_env():
-    db_user = get_environ_or_bust('POSTGRES_USER')
-    db_password = get_environ_or_bust('POSTGRES_PASSWORD')
-    db_name = get_environ_or_bust('POSTGRES_DB')
-    db_host = get_environ_or_bust('POSTGRES_HOST')
-    db_port = os.environ.get('POSTGRES_PORT', 5432)
-    db_ssl_mode = os.environ.get('POSTGRES_SSL_MODE', None)
-    db_root_cert = os.environ.get('POSTGRES_SSL_ROOT_CERTIFICATE', None)
+def create_db_logger(outputfile):
+    """Function to setup logging of SQL statements
+
+    Args:
+        outputfile (filepath): The filepath where the SQL statements will be
+            logged
+    """
+    if not outputfile:
+        return
+
+    db_log_file_name = outputfile
+    db_handler_log_level = logging.INFO
+    db_logger_log_level = logging.DEBUG
+
+    db_handler = logging.FileHandler(db_log_file_name)
+    db_handler.setLevel(db_handler_log_level)
+
+    db_logger = logging.getLogger('sqlalchemy')
+    db_logger.addHandler(db_handler)
+    db_logger.setLevel(db_logger_log_level)
+
+
+def _create_postgres_url(db_user, db_password, db_name, db_host,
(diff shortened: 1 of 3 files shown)


pint_server/database.py
```diff
@@ -39,21 +39,59 @@
...
+        return
+
+    db_log_file_name = outputfile
+    db_handler_log_level = logging.INFO
+    db_logger_log_level = logging.DEBUG
+
+    db_handler = logging.FileHandler(db_log_file_name)
+    db_handler.setLevel(db_handler_log_level)
+
+    db_logger = logging.getLogger('sqlalchemy')
```
guangyee, line 65: 'sqlalchemy.engine' right?

pint_server/database.py
```diff
@@ -39,21 +39,59 @@
...
+    Args:
+        outputfile (filepath): The filepath where the SQL statements will be
+            logged
+    """
+    if not outputfile:
+        return
+
+    db_log_file_name = outputfile
+    db_handler_log_level = logging.INFO
+    db_logger_log_level = logging.DEBUG
```
guangyee, line 60: why are the log levels hardcoded?

pint_server/models.py
```diff
@@ -42,6 +42,11 @@ class ServerType(enum.Enum):
 
 
 class PintBase(object):
+    @property
+    def tablename(self):
```
guangyee, line 46: where's this being used?

guangyee (commented):
Changes looks fine. Just some questions and bitchings.

pint_server/models.py
rtamalin: In pint-data-manager as a generic property to determine the table name without having to access a double underscore attribute outside the pint_server code base ;-)

pint_server/database.py
rtamalin: Because this log file is used to capture specific logs in an "additional" special log file to satisfy the requirement to log the SQL statements/processing independently of the standard logging.

pint_server/database.py
rtamalin: @KeithMnemonic would have to answer the specifics of that one, but my understanding was that the selections were made to ensure that we capture the specific requested data as log entries.

pint_server/database.py
```diff
@@ -39,21 +39,59 @@
...
 
-def create_postgres_url_from_env():
-    db_user = get_environ_or_bust('POSTGRES_USER')
-    db_password = get_environ_or_bust('POSTGRES_PASSWORD')
-    db_name = get_environ_or_bust('POSTGRES_DB')
-    db_host = get_environ_or_bust('POSTGRES_HOST')
-    db_port = os.environ.get('POSTGRES_PORT', 5432)
-    db_ssl_mode = os.environ.get('POSTGRES_SSL_MODE', None)
-    db_root_cert = os.environ.get('POSTGRES_SSL_ROOT_CERTIFICATE', None)
+def create_db_logger(outputfile):
```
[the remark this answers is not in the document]
rtamalin: So the purpose of this, per my understanding of the requirements, is that we are requested to collect specific additional logs as a separate log file, tracking the specific SQL actions being performed independently of the standard log.

guangyee (approved):
good with me

PR #58, opened 2021-06-17
add an option to log SQL statements on upgrade

guangyee:
Also, database.py needs to import logging before using it.


rtamalin (approved):
LGTM

PR #60, opened 2021-06-17
use psycopg2-binary instead of psycopy2

guangyee:
We don't want to recomplike the PostgreSQL DB driver on install as it
requires both gcc and postgresql12-server-devel packages.


mbelur (approved):
LGTM!

bear454:
No errors creating VENV on SLES 15 SP2 host.

PR #63, opened 2021-06-22
Add ID 

### LINKED ISSUE #75: The servers table needs a uniqueId to use as a primary key
There is no logical combination of fields to use as primary keys for servers (regions + IP may be an option), as such a new unique ID field should be added for server tables.

jesusbv:
- Remove IP and Region as PK
- Keep region non nullable
- Add ID column (integer type)
- Add revision hash
    

This Fixes #75


KeithMnemonic:
I also set @guangyee , @rtamalin and @mbelur  as reviewers as they are the ones who should also review this ideally.

rtamalin:
So first off the Pint NG code base is currently in code freeze for the Pint NG go live that is due to happen this week (today even) so no changes can be landed unless required to support that.

Secondly this is a change to the DB schema, and to support the go live, we have effectively frozen the existing schema as the "initial release" version, and any future changes to the schema will need to follow the https://github.com/SUSE-Enceladus/public-cloud-info-service#how-to-upgrade-schema workflow so that they can be added as incremental changes to the existing schema that can be migrated forward (upgrade entry point) or rolled back (downgrade entry point) using a sequence of schema versioning scripts.

pint_server/models.py
```diff
@@ -76,7 +76,7 @@ class ProviderServerBase(PintBase):
     # we decided to support other vendors, we'll need to update this
     # column type accordingly.
     ip = Column(postgresql.INET, primary_key=True)
-    region = Column(String(100), primary_key=True)
+    region = Column(String(100))
```
rtamalin, line 79: As per my comment in the main conversation, you will need to re-work this patch per the https://github.com/SUSE-Enceladus/public-cloud-info-service#how-to-upgrade-schema workflow to be able to land a schema change.

rtamalin (changes requested):

pint_server/models.py
guangyee: Yeah, this change needs to be done as part of schema migration.

jesusbv:
Before this migration
before_migration

after running the migration
after
and content of the table
select_after

same changes for the other server tables

jesusbv:
@smarlowucf for the concerns about `Enum` type or others

Adding this to the model
enum_in_migration

produces
autogenerated_enum

so it works fine

pint_server/pint_db_migrate/env.py
```diff
@@ -0,0 +1,81 @@
...
+#fileConfig(config.config_file_name)
+
+
+
+
+# add your model's MetaData object here
+# for 'autogenerate' support
+# from myapp import mymodel
+# target_metadata = mymodel.Base.metadata
+target_metadata = AmazonServersModel.metadata
```
guangyee, line 24: We should be use Base.metadata here.

```
from pint_server.models import Base
target_metadata = Base.metadata
```

pint_server/pint_db_migrate/versions/b381843009e0_trying_enums.py
```diff
@@ -0,0 +1,54 @@
...
+branch_labels = None
+depends_on = None
+
+tablenames = ['amazonservers', 'googleservers', 'microsoftservers']
+
+def upgrade():
+    # manual commands
+    op.execute("CREATE EXTENSION IF NOT EXISTS \"uuid-ossp\";")
+    for tablename in tablenames:
+        op.execute("alter table {} drop constraint {}_pkey;".format(tablename, tablename))
```
guangyee, line 24: suggest using

`op.drop_constraint('%s_pkey' % (tablename), tablename, type_='primary')`

pint_server/pint_db_migrate/env.py
jesusbv: It doesn't matter which model we use, once the PR with the `Base` one gets merged, I will solve the conflict with `Base` as well

pint_server/pint_db_migrate/versions/b381843009e0_trying_enums.py
```diff
@@ -0,0 +1,54 @@
...
+    op.execute("CREATE EXTENSION IF NOT EXISTS \"uuid-ossp\";")
```
guangyee, line 25: why are we creating the extension twice? There's one in line 22.

jesusbv: ah good catch !

pint_server/pint_db_migrate/versions/b381843009e0_trying_enums.py
```diff
@@ -0,0 +1,54 @@
...
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('amazonservers', sa.Column('id', postgresql.UUID(as_uuid=True), server_default=sa.text('uuid_generate_v4()'), nullable=False))
+    op.add_column('googleservers', sa.Column('id', postgresql.UUID(as_uuid=True), server_default=sa.text('uuid_generate_v4()'), nullable=False))
+    op.add_column('microsoftservers', sa.Column('id', postgresql.UUID(as_uuid=True), server_default=sa.text('uuid_generate_v4()'), nullable=False))
+    # ### end Alembic commands ###
+    # manual commands
+    for tablename in tablenames:
+        op.execute("alter table {} add primary key (id);".format(tablename))
```
guangyee, line 33: suggest using `op.create_primary_key()`

pint_server/pint_db_migrate/versions/b381843009e0_trying_enums.py
```diff
@@ -0,0 +1,54 @@
...
+    # ### end manual commands
+
+
+def downgrade():
+    # manual commands
+    for tablename in tablenames:
+        op.execute("alter table {} drop constraint {}_pkey;".format(tablename, tablename))
```
guangyee, line 40: suggest using `op.drop_constraint()`

pint_server/pint_db_migrate/versions/b381843009e0_trying_enums.py
```diff
@@ -0,0 +1,54 @@
...
+    op.drop_column('amazonservers', 'id')
+    op.create_table('migrate_version',
+    sa.Column('repository_id', sa.VARCHAR(length=250), autoincrement=False, nullable=False),
+    sa.Column('repository_path', sa.TEXT(), autoincrement=False, nullable=True),
+    sa.Column('version', sa.INTEGER(), autoincrement=False, nullable=True),
+    sa.PrimaryKeyConstraint('repository_id', name='migrate_version_pkey')
+    )
+    # manual commands
+    for tablename in tablenames:
+        op.execute("alter table {} add primary key (ip, region);".format(tablename))
```
guangyee, line 53: suggest using `op.create_primary_key()`

pint_server/pint_db_migrate/versions/b381843009e0_trying_enums.py
jesusbv: is there any difference/benefit in this case between the orm method vs raw SQL ?

guangyee: ORM is supposed to be database agnostic. But we are using postgresql data types directly so it's probably a moot point. But still, we should try to avoid raw SQL if we can.

jesusbv:
Update ID to integer
image
and structure updated accordingly based on confluence page
image

pint_server/pint_db_migrate/versions/b381843009e0_trying_enums.py
```diff
@@ -0,0 +1,73 @@
...
+
+def upgrade():
+    # manual commands
+    for tablename in tablenames:
+        op.drop_constraint('%s_pkey' % (tablename), tablename, type_='primary')
+        op.execute(CreateSequence(Sequence('%s_id_seq' % (tablename))))
+
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('amazonservers', sa.Column(
+        'id', sa.Integer(), nullable=False, server_default=sa.text("nextval('amazonservers_id_seq'::regclass)"))
```
jesusbv, line 29: The server default could be done in the model, but then it should be one `id` column per server model

pint_server/pint_db_migrate/versions/b381843009e0_trying_enums.py
```diff
@@ -0,0 +1,73 @@
+"""Trying Enums
+
+Revision ID: b381843009e0
+Revises: 528cb85d6ad3
```
jesusbv, line 4: Keep in mind updating this if merged in a different order

pint_server/pint_db_migrate/versions/b381843009e0_add_servers_id.py
```diff
@@ -0,0 +1,73 @@
+"""Add Servers ID
+
+Revision ID: b381843009e0
+Revises: 528cb85d6ad3
```
jesusbv, line 4: Keep in mind updating this if merged in a different order

pint_server/pint_db_migrate/versions/b381843009e0_add_servers_id.py
```diff
@@ -0,0 +1,74 @@
...
+
+def upgrade():
+    # manual commands
+    for tablename in tablenames:
+        op.drop_constraint('%s_pkey' % (tablename), tablename, type_='primary')
+        op.execute(CreateSequence(Sequence('%s_id_seq' % (tablename))))
+
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('amazonservers', sa.Column(
+        'id', sa.Integer(), nullable=False, server_default=sa.text("nextval('amazonservers_id_seq'::regclass)"))
```
rtamalin, line 29: For reference this construct doesn't appear to be necessary, and may have given me issues when testing with a postgres 12.8 based DB.

The pattern I found to work, mapped to your code, is:
```
    seq = Sequence('%s_id_seq' % (tablename))
    op.execute(CreateSequence(seq))
    op.add_column(tablename, sa.Column('id', sa.Integer, nullable=False,
                  server_default=seq.next_value())
```

PR #65, opened 2021-06-22
Add an API to return the package version

guangyee:
Add an API to return the package version of the code to improve
deployability. The new API is

GET /package-version

and it will return something like this

{
  "package version": "2.0.3"
}

### DIFF
--- a/SUSE.rst
+++ b/SUSE.rst
@@ -61,7 +61,7 @@ Testing Pint Server Lamba Function Container Locally
 How To Make A Release
 =====================
 
-1. update setup.cfg with the new version
+1. update pint_server/__init__.py '__VERSION__' attribute with the new version
 
 2. create a git annotated tag for the release version. For example:
 
--- a/pint_server/__init__.py
+++ b/pint_server/__init__.py
@@ -15,6 +15,9 @@
 # To contact SUSE about this file by physical or electronic mail,
 # you may find current contact information at www.suse.com
 
+# NOTE(gyee): must update the version here on a new release
+__VERSION__ = '2.0.3'
+
 from pint_server.database import init_db
 from pint_server.models import (AlibabaImagesModel, AmazonImagesModel,
                     AmazonServersModel, GoogleImagesModel,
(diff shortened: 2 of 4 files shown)


rtamalin:
Looks good, but probably should integrate bumpversion into the workflow to take care of incrementing the version value in an automated fashion. Potentially could even integrated it with github actions such that it automatically runs as part of merging a PR?

guangyee:
> Looks good, but probably should integrate bumpversion into the workflow to take care of incrementing the version value in an automated fashion. Potentially could even integrated it with github actions such that it automatically runs as part of merging a PR?

I would love to use bumpversion if not for IBS. IBS builds have no access to git and git tags. So it it is much easier to just manually update that one line right now. Otherwise, we'll be getting in a lots-of-pain-with-little-gain situation. :-)

mbelur (approved):
LGTM. We have an enhancement ticket for the bumpversion workflow.

rtamalin (approved):
LGTM

rtamalin:
> > Looks good, but probably should integrate bumpversion into the workflow to take care of incrementing the version value in an automated fashion. Potentially could even integrated it with github actions such that it automatically runs as part of merging a PR?
> 
> I would love to use bumpversion if not for IBS. IBS builds have no access to git and git tags. So it it is much easier to just manually update that one line right now. Otherwise, we'll be getting in a lots-of-pain-with-little-gain situation. :-)

So I'm not sure why IBS is a reason not to use bumpversion. ;-)

Using bumpversion would automate updating the version in the code, eliminating risk of human error.

If IBS needs a tag to be set, well that can be automated to; you could write a script to do the tagging for you, looking at the version value in the file associated with the SHA you want to tag to determine the version to use when taggging. And if you have a working script to do that, you could integrate it into a CI pipeline to automate the tagging process whenever a new commit is merged?

KeithMnemonic (commented):
Is there a versioning scheme? I.e when do you flip the first or second digit?

guangyee:
> Is there a versioning scheme? I.e when do you flip the first or second digit?

We don't have an official versioning scheme AFAICT. Usually, major version is reserved for features or APIs which may not be backward compatible. Anything minor is expected to be backward compatible. Releases are snapshots during a release cycle.

PR #70, opened 2021-07-27
Implement a check mode

guangyee:
Implemented a "--check-mode" option for the upgrade command. If specified,
pint_db_migrate will simply perform data sanity check (i.e. duplicates)
and skip the actual data import. If the option is not specified,
duplicates will be ignored, with a warning message posted on the screen,
but the data import will proceed.

This PR also account for any empty fields during data import.


jeremy-moffitt:
I'm confused why this was closed... does it not check the dupes correctly or was this from the RC conversation about whether this fix is needed?

guangyee:
> I'm confused why this was closed... does it not check the dupes correctly or was this from the RC conversation about whether this fix is needed?

Yes, based on RC conversation. Per my understanding, there's no chance for XML duplicates due to the validatedata script. I can reopen it if needed.

jeremy-moffitt:
I think we do want this... its lower priority than fixing the empty field update issue, but still should be on the to-do list to avoid corner case situations in deploying greenfield test instances

guangyee:
Don't merge this one yet. We're still debating the fine details.

rjschwei (approved):

rjschwei:
In general it is, IMHO, the responsibility of the data provider to ensure that there are no duplicates. In our case this is handled in the consistency checkers for the XML data currently used. That said having an independent sanity check is a good idea.

PR #71, opened 2021-07-27
Do not skip attributes that has no value

guangyee:
Those need to be accounted for.

### DIFF
--- a/bin/pint_db.py
+++ b/bin/pint_db.py
@@ -90,8 +90,10 @@ def extract_provider_data_rows(parent_node, child_name):
         row = {}
         for attr, value in child_node.items():
             if not value:
-                continue
-            if attr == 'type':
+                # NOTE: setting it to None seem to be compatible with any data
+                # type
+                attr_value = None
+            elif attr == 'type':
                 if 'smt' in value:
                     attr_value = ServerType.update
                 else:


bin/pint_db.py
```diff
@@ -90,6 +90,7 @@ def extract_provider_data_rows(parent_node, child_name):
         row = {}
         for attr, value in child_node.items():
             if not value:
+                row[attr] = None
```
smarlowucf, line 93: I think we need to be careful here. Some values are expected to be "Date" types. I think None works correctly there. However, some are expected to be strings and we should make sure that they are okay as None instead of empty strings ''.

Also, it seems like it would be better to follow the workflow of the rest of the function here. E.g. instead of continue, set `attr_value = None` (or '') and update the next case to be `elif attr == 'type':`. I think this will make the function more readable.

smarlowucf (commented):
Thanks @guangyee, a couple comments

bin/pint_db.py
guangyee: I tested this change locally. Looks like it works for any type. Let me update the if statement to make it more logical to read. Thanks for the suggestion.

smarlowucf: The cases can still be simplified.

```
if not value:
    # NOTE: setting it to None seem to be compatible with any data
    # type
    row[attr] = None
else:
    if attr == 'type':
```

Can be:

```

if not value:
    # NOTE: setting it to None seem to be compatible with any data
    # type
    attr_value = None
elif attr == 'type':
```

Which prevents increasing tab level

guangyee: We need the big else statement because the value is assigned in line #115. Or else we'll need to move that assignment into every single elif statement. So I picked my poison. :-)

smarlowucf: Yeah, if you set `row[attr]` to None then you need the outer conditional. However, if you set `attr_value` to None then the conditional acts like every other condition. Then in the end `row[attr]` gets set to `attr_value`.

guangyee: ahh, I see your point. Let me update. Thanks!

smarlowucf: I don't know if I'm explaining it well so the full code for the suggestion is:

```
if not value:
    # NOTE: setting it to None seem to be compatible with any data
    # type
    attr_value = None
elif attr == 'type':
    if 'smt' in value:
        attr_value = ServerType.update
    else:
        attr_value = ServerType.region
    # NOTE: make sure we also populate the 'shape' column if
    # 'smt' or 'regionserver' has '-sles' or '-sap' postfix.
    m = server_type_matcher.match(value)
    if m and m.group(1):
        row['shape'] = m.group(1)
    else:
        row['shape'] = ''
elif attr == 'state':
    attr_value = getattr(ImageState, value)
elif attr in ['deletedon', 'deprecatedon', 'publishedon']:
    attr_value = datetime.strptime(value, "%Y%m%d").date()
else:
    attr_value = value
row[attr] = attr_value
```

bear454 (approved):
checked out to AV3 and tested on staging server; looks good to me!

smarlowucf:
Also, LGTM. For some reason I cannot change my review to approval.

jeremy-moffitt (approved):

bear454:
Can we merge this now?

PR #72, opened 2021-07-29
switch Alembic for schema migration

guangyee:
This patch also does the following:

1. do not auto create the tables in ORM so we can detect any missing schema


pint_server/pint_db_migrate/versions/346dbf517c0e_in_the_beginning.py
```diff
@@ -0,0 +1,156 @@
...
+    # create the enums and tables if they don't exist already
+    for obj in [server_type, image_state, alibabaimages, amazonservers,
+                amazonimages, googleimages, googleservers, microsoftimages,
+                microsoftregionmap, microsoftservers, oracleimages,
+                versions]:
+        obj.create(op.get_bind(), checkfirst=True)
+
+
+def downgrade():
+    for tablename in ['alibabaimages', 'amazonimages', 'amazonservers',
```
jesusbv, line 148: can we use the table object and its `name` attribute ?

as `[amazonimages.name, alibabaimages.name,....]` 

It would be less prone to errors having to type the name once, at the creation of the table

README.rst
```diff
@@ -258,44 +258,35 @@ For example:
...
 
-     source dev_venv/bin/activate
+     cd public-cloud-info-service/pint_server
+     alembic revision -m 'fix some table'
+
+   If the above command is successful, you'll see the auto generate
+   revision file in `./pint_db_migrate/versions/`. The file is named
+   `<revision>_fix_some_table.py`.
+
+4. edit the revision file to do your thing
```
smarlowucf, line 289: This is against what was agreed upon yesterday? We don't want to be in the business of manually writing migrations.

smarlowucf (commented):
There are new migrations which seem to be manually created that lead to a drift between the models and the migrations. Which means the models which will be used by the code won't match what the database actually looks like. I thought the agreement was to autogenerate migrations so that the models and migrations are always in agreement?

README.rst
guangyee: Not sure if I understand. There's no auto-generate SQL. Even with Alembic, only thing that is auto-generated is the framework and revisions. Schema and data migrations are very specific, I have not came across a framework that will auto-generate everything for you.

pint_server/pint_db_migrate/versions/346dbf517c0e_in_the_beginning.py
guangyee: Good idea!

README.rst
smarlowucf: https://alembic.sqlalchemy.org/en/latest/autogenerate.html

guangyee: Thanks for the link. Unfortunately, there are lots of caveats spell out in the doc, which IMO quick dangerous. In our case ...

1. We use Enums and PostgreSQL-specific types such as ENUM, INET, etc which cannot be auto-detected and generated.
2. It does not automatically shrink primary keys as we intended.
3. It does not do data manipulation such as removing duplicate records prior to switching primary keys.

Auto-generate in this case is more like Tesla's autopilot feature, which doesn't do exactly as it advertised, which IMO quite dangerous. :-) We can certainly add it to README to assist user as a starting point for the migration scripts. But certainly, there will be manual editing involved.

smarlowucf: I've never had a problem with an auto generation and I've never had to manually tweak an auto generated "schema" migration. To me this PR has a big issue in that the models still don't match the migrations.

> * We use Enums and PostgreSQL-specific types such as ENUM, INET, etc which cannot be auto-detected and generated.

This is certainly opinionated (also not directly related to the migration change) but to me this is "code smell". We are using PostgreSQL but IMHO it's best to design a data model that generically works for all (most) RDBMS.

> 
> * It does not automatically shrink primary keys as we intended.

This was caused by a communication issue regarding initial design. Ideally this is not something we would be doing in migrations. See follow on question.

> 
> * It does not do data manipulation such as removing duplicate records prior to switching primary keys.

Yes, data migrations are always manual.

Question I have is, what's the difference in effort to migrate the current DB vs starting over with a flattened migration that is generated directly from the models to ensure that we keep the two pieces in sync.

guangyee: > I've never had a problem with an auto generation and I've never had to manually tweak an auto generated "schema" migration. To me this PR has a big issue in that the models still don't match the migrations.

The schema mismatch problem we had originally was due to having ORM auto-create any missing tables by default. I've
disabled that in the patch so we would never run into schema mismatch problem again. ORM will only do data update. Any missing schema will be revealed once the ORM start inserting data.

> 
> > * We use Enums and PostgreSQL-specific types such as ENUM, INET, etc which cannot be auto-detected and generated.
> 
> This is certainly opinionated (also not directly related to the migration change) but to me this is "code smell". We are using PostgreSQL but IMHO it's best to design a data model that generically works for all (most) RDBMS.

You'll have to take it up with Robert on this one. :-) That issue was raised and got shot down. Maybe you can be more convincing than I was. :-) 

> 
> > * It does not automatically shrink primary keys as we intended.
> 
> This was caused by a communication issue regarding initial design. Ideally this is not something we would be doing in migrations. See follow on question.

I disagree. Data manipulation is not something we simply cannot avoid, especially if we are to change primary keys, unique indexes, etc again in the future, unless you want to clean up your DB manually prior to schema migration. But then that would defeat the purpose.

> 
> > * It does not do data manipulation such as removing duplicate records prior to switching primary keys.
> 
> Yes, data migrations are always manual.
> 
> Question I have is, what's the difference in effort to migrate the current DB vs starting over with a flattened migration that is generated directly from the models to ensure that we keep the two pieces in sync.

Are you proposing to nuke the production DB and start over? We may need another meeting on that. :-)

smarlowucf:
Moving my biggest concern to a separate comment. Currently this PR still leaves the application level schema (models) out of sync with the db level schema (migrations).

If these changes are deployed the application will be using different primary keys then the database. Example:

migration (DB schema)
```
op.create_primary_key('microsoftimages_pkey', 'microsoftimages', ['name', 'environment'])
```

models (app schema)
```
class ProviderImageBase(PintBase):
    name = Column(String(255), primary_key=True)
    state = Column(Enum(ImageState, name=ImageState.__enum_name__))
    replacementname = Column(String(255))
    publishedon = Column(Date, primary_key=True)
    deprecatedon = Column(Date)
    deletedon = Column(Date)
    changeinfo = Column(String(255))

class MicrosoftImagesModel(Base, ProviderImageBase):
    __tablename__ = 'microsoftimages'

    environment = Column(String(50), primary_key=True)
    urn = Column(String(100))
```

guangyee:
> There are new migrations which seem to be manually created that lead to a drift between the models and the migrations. Which means the models which will be used by the code won't match what the database actually looks like. I thought the agreement was to autogenerate migrations so that the models and migrations are always in sync?



> Moving my biggest concern to a separate comment. Currently this PR still leaves the application level schema (models) out of sync with the db level schema (migrations).
> 
> If these changes are deployed the application will be using different primary keys then the database. Example:
> 
> migration (DB schema)
> 
> ```
> op.create_primary_key('microsoftimages_pkey', 'microsoftimages', ['name', 'environment'])
> ```
> 
> models (app schema)
> 
> ```
> class ProviderImageBase(PintBase):
>     name = Column(String(255), primary_key=True)
>     state = Column(Enum(ImageState, name=ImageState.__enum_name__))
>     replacementname = Column(String(255))
>     publishedon = Column(Date, primary_key=True)
>     deprecatedon = Column(Date)
>     deletedon = Column(Date)
>     changeinfo = Column(String(255))
> 
> class MicrosoftImagesModel(Base, ProviderImageBase):
>     __tablename__ = 'microsoftimages'
> 
>     environment = Column(String(50), primary_key=True)
>     urn = Column(String(100))
> ```

Good catch! Yes, we'll need to keep the models in sync. I'll fix.

README.rst
smarlowucf: > I've disabled that in the patch so we would never run into schema mismatch problem again

This PR initially led to mismatch. It's entirely possible, nothing is preventing it. The idea to use autogeneration was limit this possibility by generating migrations from models rather than manually creating migrations and hoping the models stay in sync.

> Data manipulation is not something we simply cannot avoid, especially if we are to change primary keys, unique indexes, etc again in the future
> Are you proposing to nuke the production DB and start over? We may need another meeting on that. :-)

Yeah, essentially that's the question. If the main concerns about using auto generation are because of the bandaids we are applying, and we can essentially start over with little effort, then those concerns go out the window.

pint_server/models.py
```diff
@@ -104,6 +104,7 @@ class GoogleImagesModel(Base, ProviderImageBase):
 class MicrosoftImagesModel(Base, ProviderImageBase):
     __tablename__ = 'microsoftimages'
 
+    publishedon = Column(Date)
     environment = Column(String(50), primary_key=True)
```
smarlowucf, line 108: This will assume an autogenerated PK name and thus doesn't necessarily match the migrations. We would need something like:

```
PrimaryKeyConstraint('name', 'environment', name='microsoftimages_pkey')
```

pint_server/models.py
```diff
@@ -104,6 +104,7 @@ class GoogleImagesModel(Base, ProviderImageBase):
 class MicrosoftImagesModel(Base, ProviderImageBase):
     __tablename__ = 'microsoftimages'
 
+    publishedon = Column(Date)
```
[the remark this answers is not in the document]
guangyee: Yes. I was merely doing just the one as an example so we have a reference, till we got an agreement from everyone on what the primary keys should be for all tables before finishing up the migration script.

pint_server/models.py
guangyee: Good point! I will make the change once everyone agree on the proposed primary keys in the email.

README.rst
guangyee: > This PR initially led to mismatch. It's entirely possible, nothing is preventing it. The idea to use autogeneration was limit this possibility by generating migrations from models rather than manually creating migrations and hoping the models stay in sync.
> 

I agree we should use auto-generate as a starting point. Will update README to reflect that. But sometimes we still need
edit the generated migration script if data manipulation is necessary, especially in the case where we are shrinking the primary key set.

> Yeah, essentially that's the question. If the main concerns about using auto generation are because of the bandaids we are applying, and we can essentially start over with little effort, then those concerns go out the window.

Not sure about the little effort part. That's something we'll need to work with Joel's team to come up with an estimate as this is production DB. Even if we are starting over, what are we gaining? It's still schema migration from this point onward every time we make a change. Are you saying we will never run into data problems in the future? ever? And the schema we have right now is rock solid and we'll never need to change the constraints again?

guangyee: Also, I just tried the auto generate feature, it does NOT do what we wanted. When I updated the model to remove `publishedon` from the primary keys list, the generated code does not attempt to drop the original primary key constraint, instead, it simply trying to update `nullable=True`.

```
    # ### commands auto generated by Alembic - please adjust! ###
    op.alter_column('alibabaimages', 'publishedon',
               existing_type=sa.DATE(),
               nullable=True)
    op.alter_column('alibabaimages', 'id',
               existing_type=sa.VARCHAR(length=100),
               nullable=True)
    op.alter_column('alibabaimages', 'region',
               existing_type=sa.VARCHAR(length=100),
               nullable=False)
```
This will result in errors like this when attempt to migrate.

```
Failed to upgrade Pint database: (psycopg2.errors.InvalidTableDefinition) column "publishedon" is in a primary key

[SQL: ALTER TABLE alibabaimages ALTER COLUMN publishedon DROP NOT NULL]
(Background on this error at: http://sqlalche.me/e/13/f405)
```

Bottom line, we can't trust the auto-generate code even for things such as removing a primary key.

pint_server/models.py
guangyee: Latest patchset has the proposed changes for the image tables.

pint_server/models.py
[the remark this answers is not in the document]
guangyee: Latest patchset has the proposed changes for the image tables.

README.rst
smarlowucf: Right, for certain things it does not pick them up automagically. This is to be expected and is documented. For example it's tough to know if a column is renamed or if one is deleted and a different one is added. Alembic makes it clear in the generated migration:

`# ### commands auto generated by Alembic - please adjust! ###`

Just because it might miss some types of changes doesn't mean it cannot be used. Instead I think it means that the migration would be reviewed and updated accordingly if necessary.

guangyee: For the vanilla operations like adding/removing tables/columns this works fine. But anything more than that, like changing indexes, definitely required extra work. But in any case, we should use the auto-generated script as a starting point and go from there.

pint_server/alembic.ini
```diff
@@ -0,0 +1,89 @@
+# A generic, single database configuration.
```
rtamalin, line 1: Since this file needs to be customised for people to be able to use alembic, but we don't want people to commit any changes they make to the file, perhaps we should switch to a workflow like the following:

1) We maintain an alembic.ini.example in the git repo, with an explicit .gitignore entry to ignore the alembic.ini
2) We document that people should copy the alembic.ini.example to alembic.ini and customise it as needed (this significantly reduces the change that someone inadvertently commits their creds)

Potentially we can have any dev-env setup scripting do the copy operation for us...

README.rst
```diff
@@ -258,46 +258,62 @@ For example:
...
+
+     source dev_venv/bin/activate
 
-2. create the Python 3.6 development virtual environment
+3. update `pint_server/models.py` to reflect the latest changes
+
+4. if schema change is pretty vanilla and straight forward (i.e.
+   add a new table and add/remove non-indexing columns) you can just let
+   Alembic generate the change script for you. Otherwise, skip to the next
+   step. To do that, you first need to edit `pint_server/alembic.ini` file to
```
rtamalin, line 283: See my comments below about instead providing a `pint_server/alembic.ini.example` that should be copied to `pint_server/alembic.ini`, which itself can be a .gitignore'd file, to aid in avoiding concerns about people committing credentials in PRs..

Also the `sqlalchemy.url` setting should have some pre-amble comments explaining the construction of the setting value.

README.rst
```diff
@@ -258,46 +258,62 @@ For example:
...
-     ./bin/create_dev_venv.sh
+   cd public-cloud-info-service/pint_server
+   alembic revision --autogenerate -m 'add some table'
+
+   If the above command is successful, you'll see the auto generate
+   revision file in `./pint_db_migrate/versions/`. The file is named
+   `<revision>_add_some_table.py`.
 
-3. activate the development virtual environment
+   **NOTE:** don't forget to back out your changes to `pint_server/alembic.ini`
```
rtamalin, line 297: See above and below comments about providing a reference alembic.ini.exaple that needs to be copied and customised; that represents a better approach in the long run for avoidiung risk of people committing their creds in a PR...

bin/pint_db.py
```diff
@@ -282,7 +281,7 @@ def orm_update_tables(db, provider, tables, version):
 
 
 def orm_load_database(pint_data, db_logfile=None):
-    db = init_db(outputfile=db_logfile)
+    db = init_db(outputfile=db_logfile, create_all=False)
```
rtamalin, line 284: Question: Should the default not now be to not create the tables, which the requirement being that any tooling that wants to skip the versioned DB setup is the one to specify create_all=True when calling init_db().

The default should probably be based upon the production scenario, in which case the DB should be being initialised using this pint_db.py tool, and the application side of things should be starting up pointing at an already initialised DB, and therefore shouldn't need to create any tables...

pint_server/database.py
```diff
@@ -154,7 +154,7 @@ def create_postgres_url_from_env():
 
 
 def init_db(dbconfig=None, outputfile=None, echo=None,
-            hide_parameters=None):
+            hide_parameters=None, create_all=True):
```
rtamalin, line 157: As per my comments above, perhaps this should default to false rather than true? The production workflow should only ever be starting the application side of things against an already initialised DB, and shouldn't need to create any tables...

pint_server/models.py
```diff
@@ -81,24 +82,26 @@ class ProviderServerBase(PintBase):
 
 class AmazonImagesModel(Base, ProviderImageBase):
     __tablename__ = 'amazonimages'
+    #__table_args__ = (UniqueConstraint('id', name='amazonimages_uniq_id'),)
```
rtamalin, line 85: Should this commented out line be removed? Or does it serve some "documentation" purpose? If the latter then perhaps include a preamble comment with relevant details.

requirements.txt
```diff
@@ -9,15 +10,16 @@ itsdangerous==1.1.0
 Jinja2==2.11.3
 jmespath==0.10.0
 lxml==4.6.3
+Mako==1.1.4
 MarkupSafe==1.1.1
 pbr==5.6.0
-psycopg2-binary
+psycopg2-binary==2.9.1
 python-dateutil==2.8.1
+python-editor==1.0.4
```
rtamalin, line 18: Is this a dependency of alembic?

requirements.txt
```diff
@@ -9,15 +10,16 @@ itsdangerous==1.1.0
 Jinja2==2.11.3
 jmespath==0.10.0
 lxml==4.6.3
+Mako==1.1.4
```
rtamalin, line 13: Is this a dependency of alembic?

rtamalin (changes requested):
Some questions raised, and suggestions for a better workflow to avoid risk of committing creds into a PR.

Also some lower priority suggestions for making the alembic.ini content more "readable".

README.rst
guangyee: Good idea!

bin/pint_db.py
guangyee: I don't think the ORM should be creating tables as it doesn't do sane schema migration. That's the job of the schema migration layer. The reason I set the default to `True` was to preserve backward compatibility. Even in production scenario, schema creation and data population should come before the app.

pint_server/alembic.ini
```diff
@@ -0,0 +1,89 @@
...
+
+# sys.path path, will be prepended to sys.path if present.
+# defaults to the current working directory.
+prepend_sys_path = .
+
+# timezone to use when rendering the date
+# within the migration file as well as the filename.
+# string value is passed to dateutil.tz.gettz()
+# leave blank for localtime
+# timezone =
```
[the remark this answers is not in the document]
guangyee: This file was auto-generated by `alembic init`. I just left it as is.

pint_server/database.py
guangyee: The only reason I default it to `True` was to preserve backward compatibility. If there's no such concern then we should default it to False.

pint_server/models.py
guangyee: I can remove that code. As it turned out the `id` field in `amazonimages` table is already unique. I commented it out during testing but forgot to remove it.

requirements.txt
guangyee: Yes

requirements.txt
guangyee: Yes

pint_server/alembic.ini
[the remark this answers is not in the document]
rtamalin: Ok so the alembic ini file generator sucks then ;-) No worries we can live with it...

README.rst
guangyee: Updated README.rst to always use auto generate as a starting point.

pint_server/alembic.ini
guangyee: Done.

README.rst
guangyee: Done

pint_server/models.py
guangyee: Done

pint_server/database.py
guangyee: Default to False.

rtamalin (approved):
LGTM

rtamalin:
@smarlowucf Are you happy with the latest version?

smarlowucf (commented):
Also, avoiding comments about indexes here. Will follow up to email.

README.rst
```diff
@@ -258,46 +258,54 @@ For example:
...
+
+5. uncomment and set the `sqlalchemy.url` property in
+   `pint_server/alembic.ini` to point to database to which to generate the
+   next version of the schema. Make sure the database scheme is up-to-date
+   prior to generate the next revision.
 
-3. activate the development virtual environment
+6. auto generate the next revision. Note that Alembic will use the existing
+   database as the baseline to generate the next revision so make sure the
+   existing database is up-to-date. To auto generate the next revision:
```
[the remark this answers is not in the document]
guangyee: Yes, these step works in our dev environment. We have been using the pint-devel vagrant environment (https://gitlab.suse.de/yeey/pint-devel) for all our Pint development work. There's no need to run scheme generation against a production database.

pint_server/models.py
```diff
@@ -82,23 +83,24 @@ class ProviderServerBase(PintBase):
 class AmazonImagesModel(Base, ProviderImageBase):
     __tablename__ = 'amazonimages'
 
-    id = Column(String(100), primary_key=True)
+    id = Column(String(100))
```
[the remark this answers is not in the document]
guangyee: Yeah, it was made above my pay grade. :-) But I am happy to update the code if we change our mind on this one.

smarlowucf:
As mentioned in the last discussion I though the agreement was to split this up into smaller more scuccinct PRs? There's a lot going on here and I think the index discussion is still ongoing.

guangyee:
> As mentioned in the last discussion I though the agreement was to split this up into smaller more scuccinct PRs? There's a lot going on here and I think the index discussion is still ongoing.

The only thing I can split out is the 

> As mentioned in the last discussion I though the agreement was to split this up into smaller more scuccinct PRs? There's a lot going on here and I think the index discussion is still ongoing.

Sounds good. I can definitely split up the stuff in `pint_server/pint_db_migrate/versions/` till we settle things in the email thread. Are you OK with merging the rest as is?

rjschwei (approved):

smarlowucf (approved):

rtamalin:
Since we are all happy I'm going to merge it!

PR #76, opened 2021-07-30
Add IPv6 column

jesusbv:


pint_server/models.py
```diff
@@ -77,6 +77,7 @@ class ProviderServerBase(PintBase):
     # column type accordingly.
     ip = Column(postgresql.INET, primary_key=True)
     region = Column(String(100), primary_key=True)
+    ipv6 = Column(postgresql.INET)
```
guangyee, line 80: We've agreed on putting a unique index on ip and ipv6 right?

jesusbv: No, last thing we agree (on the meeting yesterday) was an ID column as PK, which it is already done in another PR

jesusbv: at least, not in the meeting but it makes sense to have it

jesusbv: we can't have a unique index, at least, on `ip` because it is not unique, it would probably be the same for `ipv6`

pint_server/pint_db_migrate/versions/528cb85d6ad3_add_ipv6_column.py
```diff
@@ -0,0 +1,32 @@
...
+Create Date: 2021-07-30 19:05:31.695486
+
+"""
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision = '528cb85d6ad3'
+down_revision = None
```
guangyee, line 14: This will probably need to be updated once we have all the PRs lined up. Otherwise, the changes looks good!

guangyee (approved):
Looks good! We probably need to update the `down_revision` later once we have all the PRs sorted out.

pint_server/pint_db_migrate/versions/528cb85d6ad3_add_ipv6_column.py
jesusbv: yep, I thought about that all the PRs but the first would have to keep that value on track, will update when the value for this PR is clear

jesusbv: Done.

jesusbv:
This should be merged after #72

PR #82, opened 2021-08-20
switch Alembic for schema migration

guangyee:
This patch also does the following:

1. do not auto create the tables in ORM so we can detect any missing schema
2. demarcate schema migration/upgrade and data update to satisfy the
   separation of responsibilities security requirement. Schema migration
   will be handled by bin/schema_upgrade.sh while data update will be
   handled by bin/data_update.sh.


rjschwei (approved):

pint_server/pint_db_migrate/versions/3a1b9b52bd78_drop_migrate_version_table.py
```diff
@@ -0,0 +1,31 @@
...
+
+"""
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.engine.reflection import Inspector
+
+
+# revision identifiers, used by Alembic.
+revision = '3a1b9b52bd78'
+down_revision = None
```
smarlowucf, line 15: This doesn't make sense. The migration is dropping a table but has no down revision. How does the table exist if no migrations created it? This would make it impossible to start from scratch right?

You should be able to start from an empty database but since pint_server/pint_db_migrate/versions/001_in_the_beginning.py is being deleted this doesn't seem to be the case. Am I missing something?

requirements.txt
```diff
@@ -1,3 +1,4 @@
+alembic==1.6.5
```
smarlowucf, line 1: It's not ideal to do a `pip freeze > requirements.txt` which seems to be the case here?

Problem is the requirements file should reflect only direct requirements of the package and not the entire requirements tree. Also, locking very specific versions of packages down to patch level means you won't get an patches whether or not they are related to a CVE.

Ideally we would only have direct requirements with no version lock. CI/CD would be in place to catch any regressions due to package upgrades before they are updated in production.

smarlowucf (changes requested):
Why are we wiping out the first migration? We should be able to provision a new database by running all migrations.

pint_server/pint_db_migrate/versions/3a1b9b52bd78_drop_migrate_version_table.py
guangyee: This is dropping the old `sqlalchemy-migrate` tables as part of the cleanup. Starting from scratch is not our primary scenario. Our primary scenario is to work from the existing DB, created by `sqlalchemy-migrate`.

I took out `<version>_in_the_beginning.py` per your request in the original Alembic patch because we still haven't settle on the primary keys and unique indexes. You want it to be done in the subsequent patches. I can add it back on with whatever we had so far if you like, but we absolutely need to be on the same page. Please let me know.

requirements.txt
guangyee: Even with CI/CD, we absolutely need to pin the package version when working in the Python world as there's no guarantee that newer packages won't break things. If there's a CVE in a given package, we'll need up update the affected package, and retest. Package upgrade must be done carefully, along with a full set of tests.

pint_server/pint_db_migrate/versions/3a1b9b52bd78_drop_migrate_version_table.py
smarlowucf: If we have a working database that was created from migrations then these should exist in production. The changeset has the migration being deleted. It seems it's better to leave it as is and a follow on PR would fix the primary key issues.

In our case master branch is production. IMHO master should never knowingly be broken. If we are working through PRs where we expect there to be partial commits then we want this to go into a staging branch. Since the switch to alembic isn't really production ready ideally all related changes would be staged.

> Starting from scratch is not our primary scenario.

It doesn't matter if the "primary scenario" is an existing DB. If we commit 3a1b9b52bd78_drop_migrate_version_table.py  as the only migration to production (master) then we have a broken code base. IMHO changes that make their way into master/main should be production ready and tested. From a production standpoint we really want to have the ability to create a new database from scratch based on migrations. Otherwise how do we plan to provision test or development databases?

guangyee: Per my understanding, with the elimination of XML files, there will not be a scenario where we are starting from scratch. We will always start from DB snapshot, even in dev. The current pending PRs (i.e. DB dump/restore) reflects that approach. If there's a start from scratch scenario in production then I agree we need to stash the PRs into a feature branch.

requirements.txt
smarlowucf: As a note my comment here is a suggestion only.

> Even with CI/CD, we absolutely need to pin the package version when working in the Python world as there's no guarantee that newer packages won't break things.

Locking versions is only a way to hide the problem. It's certainly possible and likely that a direct or indirect requirement will eventually have a breaking change that is not handled by the code base. However, if CI/CD is in place we should catch these as soon as they are released and could respond to them. Whether that means a version limit or a code change. For example we never locked versions for mash https://github.com/SUSE-Enceladus/mash/blob/master/.virtualenv.requirements.txt. The requirement tree for mash is massive and we've never had a breaking change to one of these requirements get passed the testing stage. The fact that we don't lock versions means we are always testing against latest code and are able to catch these things proactively instead of re-actively. As you can see with the latest mash requirements we have a version lock active on `flask-jwt-extended`. This is because when 4.0.0 was released all the API tests failed due to breaking changes. Since we haven't had time to catch the code base up to these changes yet the version lock keeps us on a supported version.

guangyee: This is turning into a religious argument. :-) My point is do we want a *stable* and *reproducible* environment, or always chasing the latest (and arguably greatest)? We release a snapshot of we we need and what we can test at that point in time. Packaging pinning is being used in OpenStack for years, that's how we been able to keep our sanity. :-)

pint_server/pint_db_migrate/versions/3a1b9b52bd78_drop_migrate_version_table.py
rjschwei: > I took out `<version>_in_the_beginning.py` per your request in the original Alembic patch because we still haven't settle on the primary keys and unique indexes. 

Yes we have, the minimal set of unique entries are the primary keys. That amounts to 1 PK everywhere except for Azure where we need `name + environment`

rjschwei: > 
> > Starting from scratch is not our primary scenario.
> 
> It doesn't matter if the "primary scenario" is an existing DB. If we commit 3a1b9b52bd78_drop_migrate_version_table.py as the only migration to production (master) then we have a broken code base. IMHO changes that make their way into master/main should be production ready and tested. From a production standpoint we really want to have the ability to create a new database from scratch based on migrations. Otherwise how do we plan to provision test or development databases?

@smarlowucf I am not certain I follow your train of thought. At present I can see where you are going, we start with XML to initialize the DB and then as we apply each migration we would end up at the "desired" state of the DB. But the XML as the basis for this process is supposed to go away. As such storing the incremental migrations is for historical purposes and valuable, but they would no longer apply. What we will have is a DB that is in production and that is the starting point, meaning apply previous migrations would not fail but would also have no effect. So I am not certain I understand what you are trying to express. Test and development would be deployed by dumping the current DB, importing it into a test DB and then running the migration against the test DB.

requirements.txt
smarlowucf: Well it was only a suggestion not an argument :). I wanted to provide a counter example to "we absolutely need" since we have multiple examples of projects that don't have pinned requirements.

pint_server/pint_db_migrate/versions/3a1b9b52bd78_drop_migrate_version_table.py
smarlowucf: Well there's a few angles was trying to get at here so I'll try to make each one more clear:

- [ ] We decided to split up the previous PR into multiple PR's. So we have multiple PR's that (correct me if I'm wrong) rely on each other in some way. If this is the case we ideally shouldn't be merging this to master because we are using master in production.

- [ ] The comment/concern about others using pint server and needing migrations I realize after the fact is a non issue. You can create a starting migration from the models to provision an empty database. But it would still be confusing to someone else to find partial migrations in the repo.

- [ ] As a suggestion and opinion I think it would be nice to start the migrations with an initial schema dump. From the perspective of our production db we can run alembic with 3a1b9b52bd78 as the base revision against the DB and this will be set in the alembic_version table. But I don't think the latter is ideal.

> As such storing the incremental migrations is for historical purposes and valuable, but they would no longer apply

Alembic keeps track of the migrations that are applied to a DB via the hash in a revision table. So it's not just for historical purposes. It also allows for reverting and for flattening migrations in the future if that becomes a need. So from the perspective of the code repo the migrations are "historical". However, from the perspective of a specific DB all migrations in the stack remain functional.

jeremy-moffitt: My understanding is that the IPv6 PR depends on this one being merged first, and we separated out PK/index definitions, for which there is not a PR yet as we do not have a clear consensus. The service should not be broken at any point though unless we miss a migration step like the extraneous protected DB table from last time this was landed. If the code looks ok we should test it on the red/B/test stack before merging it, but if there are changes pending then we should wait since we'll just have to retest the update after the code is updated.

guangyee: Just want to make sure we're all clear on this one, with the current primary use case, which is to upgrade the existing DB, this PR *will not* break things. It simply switching over to the new Alembic framework and remove the junks left behind by the old framework. Merging it into master *will not* break things. Like I said, I tested that scenario enough to feel good about it. But please feel free to beat the shit out of it in the red stack before merging, if that'll increase people's comfort level. Hack I am willing to take bets. :-)

Again, we are *not* going back to the old migration framework so reverse action is not needed on the first migration. But subsequent migrations (i.e. IPv6, uniq indexes, etc), then we want to make sure those are reversible (i.e. downgrade).

jeremy-moffitt (approved):

rtamalin (approved):
LGTM

PR #83, opened 2021-08-23
Make create-postgres-url method available

jesusbv:


rtamalin:
@jesusbv This is the correct solution for leveraging the create_postgres_url() code in other projects; that is the reason why I ported back the changes we made in our temporary copy of database.py and integrated them with the changes that had been mad in this repo, because the database.py in this repo is intended for handling setting up database access whether it is read-only by the pint-ng server, or read-write for other applications.

I suggest we re-open this PR and get the change landed - it a NULL impact change from a pint-ng server perspective, but enables integration for other tools, which we need.

rtamalin (approved):
LGTM

PR #87, opened 2021-09-06
Set type and name not null

jesusbv:
- Update Server properties type and name to not allow null values


jesusbv:
image

pint_server/pint_db_migrate/versions/b720d8b8b775_set_server_type_name_not_null.py
```diff
@@ -0,0 +1,62 @@
...
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision = 'b720d8b8b775'
+down_revision = 'b381843009e0'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
```
rtamalin, line 19: We could potentially clean this up a bit by defining the ENUM once instead of using duplicate definitions here in the op.alter_column() statements.

The auto-generated code is a little dumb in this regard. ;-)

I'm sure that under the hood SQLAlchemy/PostgreSQL are detecting equivalency but it would probably make the code look cleaner.

pint_server/pint_db_migrate/versions/b720d8b8b775_set_server_type_name_not_null.py
```diff
@@ -0,0 +1,62 @@
...
+    op.alter_column('microsoftservers', 'type',
+               existing_type=postgresql.ENUM('region', 'update', name='server_type'),
+               nullable=False)
+    op.alter_column('microsoftservers', 'name',
+               existing_type=sa.VARCHAR(length=100),
+               nullable=False)
+    # ### end Alembic commands ###
+
+
+def downgrade():
```
rtamalin, line 42: Same as above; we can define the ENUM once, and reference it in all the statements.

pint_server/pint_db_migrate/versions/b720d8b8b775_set_server_type_name_not_null.py
jesusbv: Yes, I know what you mean, and I have done some refactoring in other PRs but, as a rule thumb, if the auto-generated code by alembic does not need any change, leave like that (i.e. IPv6 PR)

rtamalin:
For reference see the sort of cleanup work I did in the alembic migration script in #89 (currently in Draft and incorrectly thinks that the #88 changes are part of it too ;-) ) in the b9fcf24cc26a_images_non_null_updates.py file.

PR #88, opened 2021-09-09
Fully define the existing schema via alembic migration versions

rtamalin:
Schema Migration Changes
========================

The existing alembic migration scripts just implementg the following:
  * remove the legacy migrate_version table from the DB schema if it
    exists
  * add the new nullable ipv6 field to the Amazon, Google and MicroSoft
    servers tables.

However, if you start from a new empty database you will not be able to
upgrade to a complete database schema using the existing alembic schema
migrations, and in fact the addition of the new ipv6 field will fail as
the related tables don't exist.

This patch adds alembic migration definitions so that the original DB
schema (pre ipv6 column addition) is created before the ipv6 column is
added, allowing the schema_upgrade.sh script to successfully migrate
a new empty database enable the pint-data to be importing using
the data_update.sh script.

The new "in_the_beginning" alembic migration step is being inserted
into the migration chain between the "drop_migration_version_table"
and the "add_ipv6_column" steps, and should work for both existing
and new DB migrations.

The "in_the_beginning" migration script has been written so as to
use common column definitions where possible.

The "add_ipv6_column" migration script has been enhanced to use a
common ipv6 column definition, and iterate over the list of table
names.

Misc Usability Changes
======================

This patch also fixes that orm_exporttopostgres script so that it is
able to initialise a new empty DB; a previous change switched the
init_db() table creation parameter to default to false but neglected
to ensure that the orm_exporttopostgres was explicitly setting the
parameter to true.

Additionally this patch enhances both the schema_upgrade.py and the
data_update.py scripts to default to INFO level logging rather than
WARNING level, and adds the new --quiet/-q option to select WARNING
level as the log level if desired. Both scripts now log the traceback
to the DEBUG stream, visible if the --debug/-d option is specified.


jesusbv (approved):
LGTM

smarlowucf (approved):

Issue #77, opened 2021-08-05
Split DB schema migration and data import


jeremy-moffitt:
The automation pipeline for image updates will not do a data import each time the schema is updated. Currently an import is done with every migration. 

Acceptance Critera:
- its possible to do a schema migration without doing a data import

rtamalin:
Technically this is possible right now, but only for existing DBs that already have schema defined. Once #88 lands we will be in a position to close this task out.

Issue #86, opened 2021-09-01
Data update script should be more verbose when not using --debug


jeremy-moffitt:
The update_data.sh script does not provide sufficient indications of on-going progress as the data is being updated, but not as noisy as --debug. 

Acceptance Criteria:
# update_data.sh should output on a reasonable cadence signs of progress

Depending on the timeframes involved, this may be a status update once per batch of updates to a table, or a time based update where every minute or so the script reports "## records have been updated". There is some flexibility in the specifics, but it should be less noisy than --debug and more verbose than the current version, which may go silent for minutes at a time.

rtamalin:
I've made a start at this in some usability enhancements I've added as part of #88

Issue #74, opened 2021-07-29
The primary keys for images need to be corrected


jeremy-moffitt:
The images table is using incorrect primary keys in some cases. The keys should be as follows:

AlibabaImages
  - name, region

AmazonImages
  - name, region

GoogleImages
  - name, project

MicrosoftImages
  - name, environment

MicrosoftRegionMap
  - environment, region

OracleImages
  - id

jeremy-moffitt:
Note, there is on-going discussion in email that may change the keys from what is specified above

rtamalin:
The changes in #88 are the underpinnings to getting this work done.

PR #89, opened 2021-09-09
Fully define the existing schema via alembic migration versions

rtamalin:
Schema Migration Changes
========================

The existing alembic migration scripts just implementg the following:
  * remove the legacy migrate_version table from the DB schema if it
    exists
  * add the new nullable ipv6 field to the Amazon, Google and MicroSoft
    servers tables.

However, if you start from a new empty database you will not be able to
upgrade to a complete database schema using the existing alembic schema
migrations, and in fact the addition of the new ipv6 field will fail as
the related tables don't exist.

This patch adds alembic migration definitions so that the original DB
schema (pre ipv6 column addition) is created before the ipv6 column is
added, allowing the schema_upgrade.sh script to successfully migrate
a new empty database enable the pint-data to be importing using
the data_update.sh script.

The new "in_the_beginning" alembic migration step is being inserted
into the migration chain between the "drop_migration_version_table"
and the "add_ipv6_column" steps, and should work for both existing
and new DB migrations.

The "in_the_beginning" migration script has been written so as to
use common column definitions where possible.

The "add_ipv6_column" migration script has been enhanced to use a
common ipv6 column definition, and iterate over the list of table
names.

Misc Usability Changes
======================

This patch also fixes that orm_exporttopostgres script so that it is
able to initialise a new empty DB; a previous change switched the
init_db() table creation parameter to default to false but neglected
to ensure that the orm_exporttopostgres was explicitly setting the
parameter to true.

Additionally this patch enhances both the schema_upgrade.py and the
data_update.py scripts to default to INFO level logging rather than
WARNING level, and adds the new --quiet/-q option to select WARNING
level as the log level if desired. Both scripts now log the traceback
to the DEBUG stream, visible if the --debug/-d option is specified.

Closes: #77
Relates: #86 #74


pint_server/app.py
```diff
@@ -117,39 +131,51 @@ def json_to_xml(json_obj, collection_name, element_name):
...
+            # construct a list of elements that will make up the type value
+            type_info = [REGIONSERVER_SMT_REVERSED_MAP[server_type.value]]
+            if value:
+                type_info.append(value)
+
+            # join the elements that make up the type value with '-' to
+            # set the returned type value
+            obj_dict['type'] = '-'.join(type_info)
+
+            # TODO(fmccarthy): Do we need this???
```
rjschwei, line 161: Yes. The return for a region server looks like this:

```
<server ip="54.253.118.149" name="" region="ap-southeast-2" type="regionserver-sles"/>
```
So we either have the column in the DB and empty or we add the code to the server itself. The latter would put more special conditions in the server code, which we probably want to avoid.

rjschwei (approved):

smarlowucf (approved):

jesusbv (approved):
LGTM

rjschwei (approved):

smarlowucf (approved):

.gitignore
```diff
@@ -13,4 +13,4 @@ dist/
 AUTHORS
 ChangeLog
 alembic.ini
-
+/.vscode
```
bear454, line 16: Just a nitpick - this isn't relevant to the PR.

bin/run_standalone.sh
```diff
@@ -30,10 +30,10 @@ if [[ "$VIRTUAL_ENV" == "" ]] ; then
   exit 1
 fi
 
-export POSTGRES_USER=snotty
-export POSTGRES_PASSWORD=MasterSlobs
-export POSTGRES_DB=postgres
-export POSTGRES_HOST=127.0.0.1
+export POSTGRES_USER=${POSTGRES_USER:-snotty}
```
bear454, line 33: Are the (oddball) default values necessary here?

pint_server/models.py
```diff
@@ -59,59 +60,74 @@ def __repr__(self):
...
     # we decided to support other vendors, we'll need to update this
     # column type accordingly.
-    ip = Column(postgresql.INET, primary_key=True)
-    region = Column(String(100), primary_key=True)
+    ip = Column(postgresql.INET)
+    region = Column(String(100), nullable=False)
     ipv6 = Column(postgresql.INET)
 
+    @validates("name")
+    def validate_name(self, key, value):
```
bear454, line 86: Just a comment on style - the `@validates("name")` specificies that we're validating the name attribute; the actual validation performed is that it's not blank on update servers.

Consider `def validate_is_not_blank` ... and using the `key` in the condition & error handler.

bear454: We could re-use that validation for deprecatedon and deletedon on Images, as well.

.gitignore
rtamalin: Not really, and I can move to a separate PR if you want ;-)

bin/run_standalone.sh
rtamalin: Those are the values that @guangyee setup initially and we have continued to maintain them.

The only reason I change these lines was to allow the script to pick up the values already configured/exported in the dev-env rather than having to edit the script if it was necessary to change them.

pint_server/models.py
rtamalin: With regard to deprecatedon and deletedon, I'm assuming that you want to relate the blankness check to the image?

As for the validate_name() validator, since the definition of validator methods is table specific, and the internal implementation is table specific, in that we are checking the self.type field to see if it has a specific value, I'm not sure that it would be worth the effort to finagle things to share an implementation with the equivalent for the images table, as it may actually make the code, which is pretty minimal/simple right now, overly complex.

As regards the choice of `validate_name` as the name for this validator method, well it is accurate, apt and succint given that we are validating the name field, and then internally the implemented check is checking for specific conditions that apply to the name field. I'm not sure that the same applies for `validate_is_not_blank` given that the name field can be blank for region servers, which would suggest a name of `validate_name_is_not_blank_for_update_servers` which is getting somewhat unwieldy.

I suppose we could implement a helper function that raises ValueError if the value is blank, that `validate_name` could call... But for that to be generic enough to be shareable between valdators it would need to either take the table, key, value and an exception message fragment as arguments so that it could construct the required exception message, or take the value and exception message itself as arguments... Which would greatly increase the complexity of what is currently relatively simple code, for not a whole lot of benefit maintainability wise.

rtamalin: Good point on using the key argument in the exception message - I'll fix that now.

pint_server/models.py
```diff
@@ -59,59 +60,74 @@ def __repr__(self):
...
+        if self.type == ServerType.update:
+            if not value:
+                raise ValueError("%s.name cannot be null/empty for an update server." % self.tablename)
```
rtamalin, line 89:
```suggestion
                raise ValueError("%s.%s cannot be null/empty for an update server." % (self.tablename, key))
```

bear454 (approved):

Issue #93, opened 2021-10-13
data_update.py incremental update broken for microsoftimages table


rtamalin:
This was a side effect of the recent primary key schema changes; the `microsoftimages` table has been switched to using a unique sequence column, rather than using a primary key composed of exist columns, so the existence checking fell back on whole row checking, which caused duplicate entries to be added.

This can be worked around by using an override set of columns to use when checking for existing entries.

PR #94, opened 2021-10-13
Fix incremental update for microsoftimages table

rtamalin:
The recent schema updates to change the primary keys associated with
some fields, and associated incremental update changes to deal with
servers tables now using a sequence column for primary keys, broke
the incremental update support for the microsoftimages table.

To fix this we need to specify an appropriate set of columns that can
be used for existing entry checks in this table, so as to ensure we
detect when an existing entry is being updated, rather than an new
entry being added.

The new IDENTITY_OVERRIDES mapping has been added, currently with a
single entry for the microsoftimages table, through more can be added
if needed, to support this fix. It can be extended to support similar
overrides for other tables should the need arrise.

Closes: #93

### DIFF
--- a/pint_server/data_update.py
+++ b/pint_server/data_update.py
@@ -193,6 +193,13 @@ def extract_data_from_file(data_file, data_store):
 
         data_store[provider]['tables'] = provider_tables
 
+# Per-table overrides to used for existing entry checks
+IDENTITY_OVERRIDES = {
+    # The microsoftimages table now uses a sequence entry as the
+    # primary key, so use the values of these columns when checking
+    # for existing entries.
+    MicrosoftImagesModel.__name__: ['name', 'environment']
+}
 
 def orm_update_table(db, provider, table_name, table_rows, version):
     if table_name == "regionmap":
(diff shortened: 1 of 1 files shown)


pint_server/data_update.py
```diff
@@ -207,20 +214,39 @@ def orm_update_table(db, provider, table_name, table_rows, version):
...
+            search_data = {k:row_data[k] for k in identity_fields}
+        else:
+            search_data = {k:v for k, v in row_data.items()
+                               if getattr(model.__table__.columns,
+                                          k).primary_key}
+
+        # if no search_data was identified, fall back on doing a whole
+        # row match search, to see if there is already an exact match
+        # present.
         if not search_data:
```
guangyee, line 243: search_data should always have something right?

guangyee (approved):
looks fine

pint_server/data_update.py
rtamalin: Nope.

If no identity overrides were specified, and no columns are marked as primary keys in the table definition, then search_data can be an empty list at this point, hence assigning row_data to it so that we can do a search for an exact match, which works fine for identifying existing servers tables entries for now, but could be an issue if we need to update any servers table entries with new data before we stop using this script...

This was basically the issue we had to deal with as the `microsoftimages` table has no columns marked as primary keys under the new schema; instead it has a unique id sequence column.

As a result search_data was empty, so we were falling back on the exact match search, finding no matches, and thus adding new entries, instead of updating the existing entry as should have been the case.

jeremy-moffitt (approved):

smarlowucf (approved):

Issue #101, opened 2021-11-04
amazon, google and microsoft servers, and microsoftimages queries return new surrogate id fields


rtamalin:
Just noticed this issue when helping Brett debug why `pint microsoft images` failed but `pint microsoft images --json` worked when using a hacked version of pint that points at local test pint-ng server instance.

After some debugging I determined that `pint` was hitting an error when rendering the XML of the id field, because it was an int, not a string or unicode.

So this represents an incompatibility between the old and new pint server implementations as a result of the primary key updates.
```
% curl -s http://pdm-dev:5000/v1/google/servers | head
{
  "servers": [
    {
      "id": 1, 
      "ip": "130.211.242.136", 
      "ipv6": "", 
      "name": "", 
      "region": "asia-east1", 
      "type": "regionserver-sles"
    }, 
```

PR #102, opened 2021-11-04
Exclude new primary key id fields from REST API request results

rtamalin:
The recent changes to add sequence based id fields to the Amazon,
Google and Microsoft servers tables, and the Microsoft images table
result in extraneous id fields appearing in the REST API request
results for these entities.

We can leverage the attr exclusion mechanism that is available in
the get_formatted_dict() helper function to avoid this problem.

Closes: #101

### DIFF
--- a/pint_server/__init__.py
+++ b/pint_server/__init__.py
@@ -16,7 +16,7 @@
 # you may find current contact information at www.suse.com
 
 # NOTE(gyee): must update the version here on a new release
-__VERSION__ = '2.0.5'
+__VERSION__ = '2.0.6'
 
 from pint_server.database import (
     init_db, create_postgres_url_from_config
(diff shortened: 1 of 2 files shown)


rjschwei (approved):

smarlowucf:
Are these functions used by PDM?

rtamalin:
> Are these functions used by PDM or it's using orm models directly?

Neither - these changes are in the pint server itself, a Python3 flask app.

The pint server and PDM share the common SQLAlchemy ORM implementation from the public-cloud-info-service repo, which the pint server flask app uses to retrieve data from the database, before manipulating the results internally to format them to emulate the original pint server results formatting in the /v1/ REST API.

KeithMnemonic (approved):
LGTM

smarlowucf:
> The pint server and PDM share the common SQLAlchemy ORM implementation

Right this is why I was curious. If it affected the "get" calls for PDM updates would be borked. Since dropping the id would make an object a "new" item.

smarlowucf (approved):

jeremy-moffitt (approved):
LGTM! Glad that the excluded attributes was already supported

bear454 (approved):

PR #105, opened 2021-11-10
changeinfo URLs should end with a /

rtamalin:
Add a validator that ensures that the changeinfo URLs end in a /.

Also add a data migration step which will make the same change to all
the existing changeinfo URLs in the database for all of the provider
images tables.

The data migration step also checks to ensure that, for all providers
that can have the same image in multiple regions, if any region has a
changeinfo URL set, then one must be set for all of the regions that
the image exists in.

TODO:
  * Add validation to ensure that adding or updating entries in the
    DB fails if it would fail the requirement that changheinfo is
    either set, or not set, for all regions that the image exists
    in.

Bump version to 2.0.7.


rtamalin:
This works for me in testing for the functionality that is implemented. Note that the data_update.py will report that it makes changes to the DB (for the XML changeinfo entries that don't end in `/`) but no net change is actually committed to the DB, because the validator automatically adds back in the trailing `/` character.

We can fix this in the data_update.py script as part of loading the XML data by add in the missing trailing `/`, so that no difference is detected...

smarlowucf (approved):

jesusbv (approved):

rtamalin:
Separated out the ends in `/` handling to #113

Issue #104, opened 2021-11-09
Add validation for date fields


KeithMnemonic:
It would be very helpful to have DB checks/constraints to ensure the following dates fields are always in chronological order

publishedon
deprecatedon
deletedon

I.e deleted must be greater than or the same as deprecated, and deprecated must be greater or the same as published.  
This also implies deletedon is greater or the same as published.

rtamalin:
So this will likely require cleanup for existing entries that don't conform to those rules, e.g.

```
postgres=# select name, state, publishedon, deprecatedon, deletedon from amazonimages where deprecatedon is not Null and deprecatedon < publishedon;
             name              |  state  | publishedon | deprecatedon | deletedon  
-------------------------------+---------+-------------+--------------+------------
 suse-sles-11-sp1-v1.02.x86_64 | deleted | 2012-07-20  | 2012-03-06   | 2012-09-06
 suse-sles-11-sp1-v1.02.x86_64 | deleted | 2012-07-20  | 2012-03-06   | 2012-09-06
 suse-sles-11-sp1-v1.02.x86_64 | deleted | 2012-07-20  | 2012-03-06   | 2012-09-06
(3 rows)
```

rtamalin:
Note that we can programmatically fix these issues in an Alembic driven data migration script; the correct deprecatedon date for a deprecated image would be the publishedon date for the replacement image, which we should be able to lookup in the same table.

However if we go this route that means we also have to ensure that data_update.py doesn't undo the fixes, or else defer landing the fixes for this issue until after we stop using data_update.py.

KeithMnemonic:
@rtamalin Remember the use case where only on copy of an image was published so there is not replacement image.  pint-data-manager will handle this as `@today` or a user supplied deprecation date

rtamalin:
@KeithMnemonic Not sure that that will impact things in any way though, as the validation we are planing should still apply, namely that publishedon <= deprecatedon <= deletedon, and if the "user" specifies a date that fails that validation for either deprecatedon or deletedon, they will get an error ;-)

rtamalin:
[...]

Issue #107, opened 2021-11-15
`schema_upgrade.py` and `data_update.py` scripts should exit with failed exit status on failure


rtamalin:
The two utility scripts, `schema_upgrade.py` and `data_upgrade.py` should exit with a failed exit status if a failure occurs.

rtamalin:
FIxed by #108

PR #108, opened 2021-11-15
Handle publishedon not yet set in data validator

rtamalin:
When validating the publishedon, deprecatedon, and deletedon fields
we need to wait until published on has been set before trying to do
any validations against it.

Since the same validation is performed when triggered for any of the
three fields, it needs to support being called for deprecatedon or
deletedon before publishedon without failing, while performing all
of the validation when called for publishedon.

Also ensure that the schema_upgrade.py and data_update.py scripts
exit with a failed exit status if a failure occurs.

Closes: #104 (again), #107

### DIFF
--- a/pint_server/models.py
+++ b/pint_server/models.py
@@ -80,17 +80,26 @@ class ProviderImageBase(PintBase):
         deprecatedon = value if key == 'deprecatedon' else self.deprecatedon
         deletedon = value if key == 'deletedon' else self.deletedon
 
-        if deprecatedon and deprecatedon < publishedon:
-            raise ValueError('Image %s invalid dates specified - '
-                             'publishedon(%s) should not be after '
-                             'deprecatedon(%s)' % (self.name,
-                             str(publishedon), str(deprecatedon)))
+        # If called for deprecatedon or deletedon before publishedon
+        # has been set we have nothing to compare against, so just
+        # fall through and accept the provided value for now.
+        # Since the validator will be triggered for all 3 fields and
+        # performs the same checks each time, even if publishedon is
+        # the last field we are called for, the validator will still
+        # fail if either deprecatedon or deletedon is not valid with
+        # respect to that publishedon value.
+        if publishedon:
+            if deprecatedon and deprecatedon < publishedon:
+                raise ValueError('Image %s invalid dates specified - '
+                                 'publishedon(%s) should not be after '
+                                 'deprecatedon(%s)' % (self.name,
+                                 str(publishedon), str(deprecatedon)))
 
-        if deletedon and deletedon < publishedon:
-            raise ValueError('Image %s invalid dates specified - '
-                             'publishedon(%s) should not be after '
-                             'deletedon(%s)' % (self.name,
-                             str(publishedon), str(deletedon)))
+            if deletedon and deletedon < publishedon:
+                raise ValueError('Image %s invalid dates specified - '
+                                 'publishedon(%s) should not be after '
+                                 'deletedon(%s)' % (self.name,
+                                 str(publishedon), str(deletedon)))
 
         if deprecatedon and deletedon and deletedon < deprecatedon:
             raise ValueError('Image %s invalid dates specified - '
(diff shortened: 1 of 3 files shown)


rtamalin:
Tested full import and incremental update cases to ensure we get a failure for a bogus publishedon date set on the first amazon image entry in XML data:

Full import:
```
% ../bin/data_update.sh --host pdm-dev --user snotty --password MasterSlobs --database postgres update
INFO:__main__:Updating data
INFO:__main__:Extracting XML data for 'alibaba' provider...
INFO:__main__:Extracting XML data for 'oracle' provider...
INFO:__main__:Extracting XML data for 'microsoft' provider...
INFO:__main__:Extracting XML data for 'amazon' provider...
INFO:__main__:Extracting XML data for 'google' provider...
INFO:__main__:Added 400 entries for model 'AlibabaImagesModel'
INFO:__main__:Adding 'versions' table entry for table 'alibabaimages' with version '20211109.0'
INFO:__main__:Added 70 entries for model 'OracleImagesModel'
INFO:__main__:Adding 'versions' table entry for table 'oracleimages' with version '20211004.0'
INFO:__main__:Added 3074 entries for model 'MicrosoftImagesModel'
INFO:__main__:Adding 'versions' table entry for table 'microsoftimages' with version '20211110.0'
INFO:__main__:Added 189 entries for model 'MicrosoftServersModel'
INFO:__main__:Adding 'versions' table entry for table 'microsoftservers' with version '20211110.0'
INFO:__main__:Added 155 entries for model 'MicrosoftRegionMapModel'
Failed to upgrade Pint database: Image suse-manager-3-2-proxy-byos-v20200506-hvm-ssd-x86_64 invalid dates specified - publishedon(2021-05-06) should not be after deprecatedon(2020-05-06)
```

Incremental update:
```
% ../bin/data_update.sh --host pdm-dev --user snotty --password MasterSlobs --database postgres update
INFO:__main__:Updating data
INFO:__main__:Extracting XML data for 'alibaba' provider...
INFO:__main__:Extracting XML data for 'oracle' provider...
INFO:__main__:Extracting XML data for 'microsoft' provider...
INFO:__main__:Extracting XML data for 'amazon' provider...
INFO:__main__:Extracting XML data for 'google' provider...
INFO:__main__:No new entries found for model 'AlibabaImagesModel'
INFO:__main__:No new entries found for model 'OracleImagesModel'
INFO:__main__:No new entries found for model 'MicrosoftImagesModel'
INFO:__main__:No new entries found for model 'MicrosoftServersModel'
INFO:__main__:No new entries found for model 'MicrosoftRegionMapModel'
Failed to upgrade Pint database: Image suse-manager-3-2-proxy-byos-v20200506-hvm-ssd-x86_64 invalid dates specified - publishedon(2021-05-06) should not be after deprecatedon(2020-05-06)
```

pint_server/models.py
```diff
@@ -80,17 +80,26 @@ def validate_image_dates(self, key, value):
...
-                             'publishedon(%s) should not be after '
-                             'deprecatedon(%s)' % (self.name,
-                             str(publishedon), str(deprecatedon)))
-
-        if deletedon and deletedon < publishedon:
-            raise ValueError('Image %s invalid dates specified - '
-                             'publishedon(%s) should not be after '
-                             'deletedon(%s)' % (self.name,
-                             str(publishedon), str(deletedon)))
+        # If called for deprecatedon or deletedon before publishedon
```
guangyee, line 83: I think we still need to check to see if there's no garbage in.

```
if type(value) != str:
    raise ValueError("Expect the value of %s to be a string, got %s instead." % (key, type(value))
```

rtamalin: Technically I believe that issue arose because the pytest setup in the pint-data-manager tests was using string date values rather than a valid datestamp type in it's mocking.

Also the above suggestion would actually break things because the values in question are not supposed to be strings at this point, but datestamps, as per the error case this PR is seeking to fix (`Failed to upgrade Pint database: '<' not supported between instances of 'datetime.date' and 'NoneType'`) so the suggested exception would in fact be triggered for valid values, rather than invalid ones.

We could instead explore leveraging the validator to coerce the provided value to be of the correct type, but I think https://docs.sqlalchemy.org/en/14/core/custom_types.html#typedecorator-recipes would be a better approach...

rjschwei (approved):

guangyee (approved):
lgtm

Issue #100, opened 2021-11-03
can add same IP address as server more than once.


brett060102:
The following sequence does not return any errors:

   61  pint-data-manager amazon servers --mode add --type update --ip-address 107.22.231.220  --region us-east-2
   62  pint-data-manager amazon servers --mode add --type update --ip-address 107.22.231.220  --region us-east-1
   63  pint-data-manager amazon servers --mode add --type update --ip-address 107.22.231.220  --region us-east-2
   64  pint-data-manager amazon servers --mode add --type region --ip-address 107.22.231.220  --region us-east-2
   65  pint-data-manager amazon servers --mode add --type region --ip-address 107.22.231.220  --region us-east-1
   66  pint-data-manager amazon servers --mode add --type region --ip-address 107.22.231.220  --region us-east-2
   67  pint-data-manager amazon servers --mode add --type region --ip-address 107.22.231.220  --region us-east-1

jeremy-moffitt:
Note that adding the same IP to multiple regions should be possible for Azure, but not for the other CSPs

rtamalin:
Turns out we sometimes temporarily add the server from another region to a newly created region in Amazon, so we need to allow for that.

This means that we need to extend the Amazon servers `ip` and `ipv6` indices to include the region field. And it makes sense to do the same thing for the Google servers table, even though we haven't needed it up to now.

This will be a modification to the previous work done in

PR #112, opened 2021-11-30
IP Addresses must be unique per region

rtamalin:
Previously we defined new partial unique indices to ensure that IP
addresses were uniquely specified in the Amazon, Google and Microsoft
servers tables, though we allowed for them being unique per region in
the case of Microsoft.

However, we sometimes temporarily specify the update servers from an
existing region as the update servers for a newly created region until
the update infrastructure for the new region is in place.

To address this operational requirement we need to include the region
field in the partial unique indices for all servers tables, not just
the Microsoft servers table.

Update the adhoc_db duplicate_servers test to reflect the fact that
all servers tables, rather than just the Microsoft servers table,
should have the second region with the repeated IP address entry
added to them.

Cleanup some comments in the report_duplicates() helper method and
extend it to take as argument the index of the field to be counted,
defaulting to the first field, rather than being hardcoded to use
the first field.

Closes: #100 (again)

### DIFF
--- a/bin/adhoc_db
+++ b/bin/adhoc_db
@@ -101,24 +101,21 @@ def duplicate_servers():
         _commit_db_changes(db, f'duplicate_servers ({table_name} add v4 '
                            'and v6 only entries)')
 
-        # Microsoft can have the same update server in multiple regions so add
-        # an update server entry in another region with both addresses
-        if table_name == 'microsoftservers':
-            LOG.debug('%s: Add other region entry', table_name)
-            row = model(**server_entry_args['v4_v6_eu'])
-            db.add(row)
+        LOG.debug('%s: Add other region entry', table_name)
+        row = model(**server_entry_args['v4_v6_eu'])
+        db.add(row)
 
-            try:
-                _commit_db_changes(db, f'duplicate_servers ({table_name} '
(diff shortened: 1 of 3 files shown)


rtamalin:
Testing against a DB seeded with the current pint-data XML (which includes the temporary duplicated server for a recently created Amazon region) content this now works, rather than failing due to detecting a duplicate:

```
% ../bin/schema_upgrade.sh --host pdm-dev --user snotty --password MasterSlobs --database postgres upgrade
INFO:__main__:Creating version control
INFO:__main__:Upgrading schema
INFO:alembic.runtime.migration:Context impl PostgresqlImpl.
INFO:alembic.runtime.migration:Will assume transactional DDL.
INFO:alembic.runtime.migration:Running upgrade 8c5fc3cd9b18 -> 14ae3b1b5e81, image dates validation
INFO:alembic.runtime.migration:Running upgrade 14ae3b1b5e81 -> e937749e3f8b, servers partial unique keys
Pint database schema migration successfully completed.
```

bear454 (approved):

rjschwei (approved):

jeremy-moffitt (approved):
LGTM

smarlowucf:
:+1:

PR #118, opened 2022-07-06
Add db_fixup script for running adhoc DB fixes

rtamalin:
Initially derived from the bin/adhoc_db test tool, and enhanced to focus on running adhoc DB fix operations.

Currently provides the following fixes:
* fix_15sp4_changeinfo: fix all 15sp4 images to have an empty changeinfo value.
* fix_leap153_ec2_state: fix specific Leap 15.3 images to correctly show as deprecated rather than deleted


rjschwei:
This should remain SUSE internal as the fixes apply only to us for adding incorrect information to the DB in the first place.

PR #120, opened 2022-07-08
Add a new route to retrieve the postgres server version

KeithMnemonic:
This PR adds a new route '/db-server-version which enables the ability to retrieve the psql server version, This is needed for pint-data-manager so it can dynamically determine the correct version of pg_dump to use. Currently pint-data-manager incorrectly is set for

`pg_dump = '/usr/lib/postgresql12/bin/pg_dump'`  

When testing upgrading the pint-server to PostgreSQL to version 14, an issue was found when running pint-data-manager due to the version mismatch between the psql server and pg_dump

Instead of just fixing this in pint-data-manager, it was suggested to make this a server function in case some other client would need access to the database server version.

### DIFF
--- a/pint_server/database.py
+++ b/pint_server/database.py
@@ -21,6 +21,7 @@
 import enum
 import logging
 import os
+import re
 
 from sqlalchemy import create_engine
 from sqlalchemy.orm import scoped_session, sessionmaker
@@ -213,3 +214,13 @@ def init_db(dbconfig=None, outputfile=None, echo=None,
         Base.metadata.create_all(bind=engine)
 
     return db_session
+
+
+def get_psql_server_version(db_session):
+    """Return the psql server version"""
+    result = db_session.execute("select version()")
+    for row in result:
+        version = re.search(r'PostgreSQL\s+\d+.\d+', str(row))
+        if version:
+            break
+    return version.group(0)
--- a/pint_server/app.py
+++ b/pint_server/app.py
@@ -29,7 +29,7 @@ from xml.dom import minidom
 import xml.etree.ElementTree as ET
 
 import pint_server
-from pint_server.database import init_db
+from pint_server.database import init_db, get_psql_server_version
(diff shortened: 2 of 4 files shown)


guangyee (approved):
looks good to me

pint_server/database.py
```diff
@@ -213,3 +214,11 @@ def init_db(dbconfig=None, outputfile=None, echo=None,
         Base.metadata.create_all(bind=engine)
 
     return db_session
+
+
+def get_psql_server_version(db_session):
+    """Return the psql server version"""
+    result = db_session.execute("select version()")
+    for row in result:
+        version = re.search('(PostgreSQL)?[0-9]+.[0-9]+', repr(row))
```
rtamalin, line 223: Few tips:
* since you are using Python `re` you can just use `\d` instead of `[0-9]`
* a regexp string should be an r-string, i.e. `r'...'` so that you don't need to do double backslashes to specify a backslash in the string

Also, is there a reason for using `repr(row)` rather than `str(row)` or just searching in `row` directly?

KeithMnemonic: using row directly does not work, i can use str.  Let me switch to r' and \d

pint_server/database.py
```diff
@@ -213,3 +214,11 @@ def init_db(dbconfig=None, outputfile=None, echo=None,
         Base.metadata.create_all(bind=engine)
 
     return db_session
+
+
+def get_psql_server_version(db_session):
+    """Return the psql server version"""
+    result = db_session.execute("select version()")
+    for row in result:
```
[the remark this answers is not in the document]
KeithMnemonic: please review the changes

pint_server/database.py
KeithMnemonic: fixed

KeithMnemonic: fixed

Issue #121, opened 2022-07-11
Query for image deletion


rjschwei:
Our image deletion date is defined as deprecation-date + 6 months. We have a user of the API that is interested in being able to query the expected deletion date. For this we have 2 use cases

1.) Receive a list of all images to be deleted on or before a given date
  + The user must be able to filter this result set by framework and region

2.) Get the expected deletion date for a given image name

rtamalin:
We identified the follow new V1 API requests:

* `/v1/images/deletedby/{date}`
* `/v1/{provider}/images/deletedby/{date}`
* `/v1/{provider}/{region}/images/deletedby/{date}`
* `/v1/{provider}/images/deletiondate/{image}`
* `/v1/{provider}/{region}/images/deletiondate/{image}`

For the provider specific deletedby API requests the result will be a list, possibly empty, of images that will be marked as deleted by the specified date, with the results looking similar to the existing provider image list queries.

For the provider specific deletiondate API requests the result with be a deletion date value, which may be empty if the specified image couldn't be found, or the image is not yet marked as deprecated.

The global images deletedby query is a little different; the result will be a list of providers, and for each provider there will be a potentially empty list of images that will be marked as deleted by the specified date. The XML result of such a query should look similar to:
```
<providers>
  <provider name="alibaba"/>
  <provider name="amazon"/>
  <provider name="google">
    <images>
      <image deletedon="".../>
      <image deletedon="".../>
    </images>
  </provider>
  <provider name="microsoft">
    <images>
      <image deletedon="".../>
      <image deletedon="".../>
    </images>
  </provider>
  <provider name="oracle"/>
</providers>
```

rtamalin:
Changes landed in #123 implement the provider specific deletedby and deletiondate requests.

The global deletedby request will be implemented in a follow on submission.

PR #123, opened 2022-07-21
Add deletion date related queries

rtamalin:
Add the following new queries (both JSON and XML):

  * Deletion date for images on a per provider basis, optionally
    scoped to a specific region.

    Will return an empty result for an image that has not yet been
    deprecated, or the expected deletion date for an image in the
    deprecated state, or the date on which an image was marked
    deleted for images in the deleted state.

  * Images that will be marked as deleted by a specific date, for
    all providers, or for a specific provider, optionally scoped
    to a specific region.

    Will return a list of the currently deprecated images that will
    be marked as deleted by the specified date.

To support these operations the dateutil.relativedelta module is
leveraged to generate an appopriate provider specific time delta
to be used in time calculations.

Deletion Date for Images Queries
================================

Two new requests have been added (both JSON and XML):

  * `/v1/<provider>/images/deletiondate/<image>`
  * `/v1/<provider>/<region>/images/deletiondate/<image>`

The <image> value should be a valid image name for the specified
provider (and optionally region).

Will return a 'deletiondate' entry with the determined deletion
date if a matching image was found that is in either the deprecated
or deleted states.

Images that will be Deleted By Queries
======================================

Three new requests have been added (both JSON and XML):

  * `/v1/images/deletedby/<date>` (disabled for now)
  * `/v1/<provider>/images/deletedby/<date>`
  * `/v1/<provider>/<region>/images/deletedby/<date>`

The <date> must be 8 decimal digits, with the first 4 digits being
the year, the next 2 digits being the month, and the final 2 digits
being the day, e.g. 20220721.

The provider specific queries will return a potentially empty list
of currently deprecated images that will be marked as deleted by
the specified <date>.

The general images query will return a hash which will contain one
entry per provider that will contain a potentially empty list of
any currently deprecated provider images that will be marked as
deleted by the specified <date>. However this request is currently
disabled due to issues with the XML response generation.

Miscellaneous Code Enhancements
===============================

As part of implementing the new queries I identified some changes
that could simplify the general code, and somewhat improve the
maintainability by converting duplicate code patterns into helper
functions.

  * Use sets instead of lists for hashable entites which handles
    duplicate elimination for free.

  * Similarly extract the common pattern of code surrounding the
    formatting of the results dictionaries for images and servers
    to simplify the request handling code.

Implements: #121

### DIFF
--- a/pint_server/app.py
+++ b/pint_server/app.py
@@ -325,42 +325,7 @@ def get_provider_servers_types(provider):
         return [{'name': 'smt'}, {'name': 'regionserver'}]
 
 
-# generate a provider specific hash that is dependent on the data
-# versions of the respective provider's category tables, if they
-# exist.
-# Optionally specify a category of table as a string or a list of
-# categories if desired. Defaults to matching images and servers
-# tables for a provider.
-def get_provider_hash(provider, categories=None):
-    if categories is None:
-        categories = ['images', 'servers']
-    elif type(categories) is str:
-        categories = [categories]
-
-    provider_tables = [provider + c
-                       for c in categories]
-
-    found_versions = (v.version
-                for v in VersionsModel.query.filter(
-                    VersionsModel.tablename.in_(provider_tables))
-                    .order_by(VersionsModel.tablename)
-    )
-
-    # include the provider name in the hashed content
-    return hash((provider, found_versions))
-
-
-# caches results of retrieving regions list for providers; cache_hash
-# argument can be used to ensure stale entries get expired. Maintains
-# at most one active cache entry for each provider. This caching works
-# for both short lived instances handling a single request, as it will
-# cache the region list for a specific provider so it only needs to be
-# calulated once, as well as for persistent server instances, even if
-# the DB data changes.
-@lru_cache(maxsize=len(PROVIDER_IMAGES_MODEL_MAP))
-def _query_provider_regions(provider, cache_hash):
-    del cache_hash  # has served it's purpose and is not used below
-
+def query_provider_regions(provider):
     if provider == 'microsoft':
         return _get_all_azure_regions()
 
@@ -383,15 +348,6 @@ def _query_provider_regions(provider, cache_hash):
             region_list.append(image.region)
 
     return region_list
-
-
-def query_provider_regions(provider):
-    # pass in a hash based on provider table data versions as
-    # one of the arguments to the lru_cache'd function to ensure
-    # that stale entries will be skipped/expired if the DB tables
-    # were updated since the last call.
-    return _query_provider_regions(provider,
-                                   get_provider_hash(provider))
 
 
 def get_provider_regions(provider):
--- a/pint_server/app.py
+++ b/pint_server/app.py
@@ -21,7 +21,6 @@ import re
 from collections import namedtuple
 from dateutil.relativedelta import relativedelta
 from decimal import Decimal
-from functools import lru_cache
 from flask import (
     abort,
     Flask,
@@ -572,9 +571,26 @@ def get_image_deletiondate_in_provider(image, provider, region=None):
                  if i.state in [ImageState.deprecated,
                                 ImageState.deleted]}
 
-    # TODO(rtamalin): is this even possible?
+    # If we get back multiple sets of results, we should use
+    # the earliest deprecatedon or deletedon date depending on
+    # the found state(s).
     if len(image_set) > 1:
-        abort(Response('', status=404))
+        image_states = {i.state for i in image_set}
+        deprecatedon_dates = sorted({i.deprecatedon for i in image_set if i.deprecatedon})
+        if not deprecatedon_dates:
+            deprecatedon_dates = ['']
+        deletedon_dates = sorted({i.deletedon for i in image_set if i.deletedon})
+        if not deletedon_dates:
+            deletedon_dates = ['']
+
+        if ImageState.deleted in image_states:
+            image_state = ImageState.deleted
+        else:
+            image_state = ImageState.deprecated
+
+        image_set = set([DeletionDetails(image_state,
+                                         deprecatedon_dates[0],
+                                         deletedon_dates[0])])
 
     # if image is in not in deprecated/deleted state the image set
     # will be empty, so the result will be an empty deletiondate.
--- a/pint_server/app.py
+++ b/pint_server/app.py
@@ -18,7 +18,10 @@
 import datetime
 import math
 import re
+from collections import namedtuple
+from dateutil.relativedelta import relativedelta
 from decimal import Decimal
+from functools import lru_cache
 from flask import (
     abort,
     Flask,
@@ -45,6 +48,32 @@ from pint_server.models import (ImageState, AmazonImagesModel,
                                 AmazonServersModel, MicrosoftServersModel,
                                 GoogleServersModel, ServerType,
                                 VersionsModel, MicrosoftRegionMapModel)
+
+
+# hashable helper class to create named tuples from the details in
+# an image entry that are relevant for deletion date tracking.
+DeletionDetails = namedtuple("DeletionDetails",
+                             " ".join(["state",
+                                       "deprecatedon",
+                                       "deletedon"]))
+
+
+# helper regexp matcher for dates specified in the %Y%m%d (4 digits
+# for year, 2 digits for month, 2 digits for day) format
+date_matcher = re.compile(
+    r'\d{4}' +  # 4 digit year
+    r'(' +  # 2 digit month, 2 digit day
+    # Months with 31 days
+    r'(01|03|05|07|08|10|12)(0[1-9]|[1-2][0-9]|3[0-1])' +
+    r'|' +
+    # Months with 30 days
+    r'(04|06|09|11)(0[1-9]|[1-2][0-9]|30)' +
+    r'|' +
+    # February
+    r'02(0[1-9]|[1-2][0-9])' +
+    r')' +
+    r''
+)
 
 
 app = Flask(__name__)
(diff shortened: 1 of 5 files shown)


rtamalin:
Note that I still need to update the functional tests to match the code changes - this is just the initial implementation of the new API requests as well as some code cleanup that happened along the way...

pint_server/app.py
```diff
@@ -212,7 +312,42 @@ def get_provider_servers_types(provider):
...
+
+
+# caches results of retrieving regions list for providers; cache_hash
+# argument can be used to ensure stale entries get expired. Maintains
+# at most one active cache entry for each provider. This caching works
+# for both short lived instances handling a single request, as it will
+# cache the region list for a specific provider so it only needs to be
+# calulated once, as well as for persistent server instances, even if
+# the DB data changes.
+@lru_cache(maxsize=len(PROVIDER_IMAGES_MODEL_MAP))
```
guangyee, line 347: I think in-memory caching will be marginally beneficial as the lambda container have a limited lifespan (of 15 minutes?) https://aws.amazon.com/lambda/faqs/.

pint_server/app.py
```diff
@@ -452,11 +688,27 @@ def assert_valid_provider(provider):
         abort(Response('', status=404))
 
 
+def assert_valid_provider_region(provider, region):
+    provider_regions = query_provider_regions(provider)
```
guangyee, line 692: This will result in 2 queries instead of one. Not sure what the performance impact is for high volume calls. I think we don't need this as the query will result in 404 anyway for an invalid region.

pint_server/app.py
```diff
@@ -38,6 +41,32 @@
...
+# an image entry that are relevant for deletion date tracking.
+DeletionDetails = namedtuple("DeletionDetails",
+                             " ".join(["state",
+                                       "deprecatedon",
+                                       "deletedon"]))
+
+
+# helper regexp matcher for dates specified in the %Y%m%d (4 digits
+# for year, 2 digits for month, 2 digits for day) format
+date_matcher = re.compile(
```
guangyee, line 54: Instead of regex, why not just let `datetime.strptime()` handle the format checking?

pint_server/app.py
```diff
@@ -126,6 +187,12 @@ def json_to_xml(json_obj, collection_name, element_name):
 
 
 def get_formatted_dict(obj, extra_attrs=None, exclude_attrs=None):
+    # make exclude attrs an empty list if not specified so that we
+    # don't have to check for it being non-None on every iteration
+    # below
+    if exclude_attrs is None:
+        exclude_attrs = []
```
guangyee, line 194: Nice!

pint_server/app.py
rtamalin: Because when I called it with, for example `111111` it comes back with year=1111, month=1, day=1... So the regex helps with ensuring we only accept a valid 8 digit date spec.

```
>>> import datetime
>>> DATE_FORMAT = '%Y%m%d'
>>> datetime.datetime.strptime('11111111', DATE_FORMAT)
datetime.datetime(1111, 11, 11, 0, 0)
>>> datetime.datetime.strptime('111111', DATE_FORMAT)
datetime.datetime(1111, 1, 1, 0, 0)
>>> datetime.datetime.strptime('1111111', DATE_FORMAT)
datetime.datetime(1111, 11, 1, 0, 0)
>>> datetime.datetime.strptime('11111111', DATE_FORMAT)
datetime.datetime(1111, 11, 11, 0, 0)
>>> datetime.datetime.strptime('11110101', DATE_FORMAT)
datetime.datetime(1111, 1, 1, 0, 0)
>>> datetime.datetime.strptime('1111101', DATE_FORMAT)
datetime.datetime(1111, 10, 1, 0, 0)
>>> datetime.datetime.strptime('1111011', DATE_FORMAT)
datetime.datetime(1111, 1, 1, 0, 0)
```

pint_server/app.py
rtamalin: So the in-memory caching means we only really do the provider regions call once for a given region, with the same data versioning hash. It has a small performance gain for an active stream of requests, and allows for moving the region validity checking to the request entry point handler, without the negative impact of having to do the actual query twice if the region specified was valid.

pint_server/app.py
rtamalin: The first query, for an empty cache, will trigger the DB operations to determine the provider specific regions list, whereas the subsequent runs will get the value from the cache.

The use of the hash argument in the @lru_cache'd function does incur additional DB activity... But it is needed to ensure that we don't incorrectly re-use stale regions data...

The Python version we are currently based on doesn't really have a decent TTL caching mechanism; newer Python releases have something more useful in that regard.

An alternative that doesn't lead to DB operations (which is what I started out with) would be to have the get_provider_hash() method just return a timestamp value rounded to the end of the nearest X second boundary window, ensuring that cache entries will expire once we move to a new boundary window...

However that approach runs the risk of inappropriately using stale cached data if for example we use a 1 minute boundary, and a query comes in at 1 second into the window, and then a DB update modifies the region details for that provider soon afterwards, say 5 seconds into the minute... Until the end of the minute boundary window, e.g. another 55 seconds, the cached data will be used which is stale...

So the hashing of the data version details is a compromise to ensure that we don't use stale data, while providing the means to cache data that might be re-used which could be a bit expensive to generate in the first place...

I'm happy to switch back the timestamp boundary window idea if this seems too expensive, but we probably will need to keep the boundary window pretty small to avoid prolonged re-use of stale cached data, which negates most of the benefit of the caching.

pint_server/app.py
rtamalin: Modern Python byte code optimisation probably eliminates any performance penalty by detecting the loop invariant and moving resolution of it to before entering the loop code, but changing the exclude_attrs to a list here removes the need to include the test in the written code, which makes it a little easier to read. ;-)

pint_server/app.py
rjschwei: The question is also how we configure the lambda function, we could end up with a cache for each provider in memory, is the lambda function configured to handle this?

pint_server/app.py
```diff
@@ -306,70 +452,178 @@ def _get_azure_images_for_region_state(region, state):
...
+                                ImageState.deleted]}
+
+    # TODO(rtamalin): is this even possible?
+    if len(image_set) > 1:
         abort(Response('', status=404))
-    exclude_attrs = PROVIDER_IMAGES_EXCLUDE_ATTRS.get(provider)
-    return [get_formatted_dict(image, exclude_attrs=exclude_attrs)
-            for image in images]
 
+    # if image is in not in deprecated/deleted state the image set
```
rjschwei, line 593: s/in//

pint_server/app.py
```diff
@@ -306,70 +452,178 @@ def _get_azure_images_for_region_state(region, state):
...
+    # helper to create a hashable object.
+    image_set = {DeletionDetails(i.state,
+                                 i.deprecatedon,
+                                 i.deletedon)
+                 for i in images
+                 if i.state in [ImageState.deprecated,
+                                ImageState.deleted]}
+
+    # TODO(rtamalin): is this even possible?
+    if len(image_set) > 1:
```
rjschwei, line 590: Well that'll depend on the client of the code, if `image` is a full image name than `> 1` cannot happen (assuming that the __hash__ method of a namedtuple produces consistently the same value for the same content). But if the content of `image` is such that multiple matches can occurr then the set could be longer than 1.

pint_server/app.py
```diff
@@ -212,7 +312,42 @@ def get_provider_servers_types(provider):
...
-def get_provider_regions(provider):
+# generate a provider specific hash that is dependent on the data
+# versions of the respective provider's category tables, if they
+# exist.
+# Optionally specify a category of table as a string or a list of
+# categories if desired. Defaults to matching images and servers
+# tables for a provider.
+def get_provider_hash(provider, categories=None):
+    if categories is None:
+        categories = ['images', 'servers']
```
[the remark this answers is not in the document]
rtamalin: See above explanation for Python default parameter best practices.

pint_server/app.py
rtamalin: My understanding is that the lambda function instance sticks around for a few mins before being disposed of, so if subsequent queries arrive during that timeframe then then could potentially benefit from the cached values. If there is concern about the size of the cached values (which isn't necessarily too large right now) we could reduce the cache size down to 1 so that the caching benefit is primarily focused the current request, and only benefits subsequent requests if they are for the same provider.

pint_server/app.py
```diff
@@ -161,14 +228,49 @@ def get_formatted_dict(obj, extra_attrs=None, exclude_attrs=None):
...
+                                        extra_attrs=extra_attrs)
+                     for r in results]
+    except DataError:
+        abort(Response('', status=404))
+
+    return formatted
+
+
+# Formatting helper for provider image list results
+def formatted_provider_images(provider, images, extra_attrs=None):
```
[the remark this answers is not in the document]
rtamalin: See above explanation for Python default parameter best practices.

pint_server/app.py
rtamalin: My understanding is that the lambda function instance sticks around for a few mins before being disposed of, so if subsequent queries arrive during that timeframe then then could potentially benefit from the cached values. If there is concern about the size of the cached values (which isn't necessarily too large right now) we could reduce the cache size down to 1 so that the caching benefit is primarily focused the current request, and only benefits subsequent requests if they are for the same provider.

pint_server/app.py
rtamalin: So right now the `query_image_in_provider_region()` call is performing an exact match search for the provided image name matching the image name in the provider's images table, so we will only ever see results for one image, and all of the results should be for that image name.

My question was more about whether it would ever be possible for the deletion related attrs used in constructing the set values, namely the `state`, `deprecatedon` and `deleted`, to be different for the same image, at least intentionally? Basically, for a provider with region specific images, for a specific image name, could we intentionally see different states, deletedon or deprecatedon values for that image?

If not then we can get rid of this check ;-)

rjschwei: In any given provider the set should always be consistent, i.e. if `suse-sles-sap-12-sp3-v20220304-hvm-ssd-x86_64` is set to deprecate in one region of the provider it should be set to the same state (deprecated) in all regions of that provider.

rjschwei (approved):

pint_server/app.py
rtamalin: FYI due to issues seen with microsoft (Blackforest environment) and AWS (some regions have had images already deprecated) we can have more than one set of results in the image_set, so we handle this by reporting deletiondate based upon the earliest possible deprecatedon or deletedon date depending on whether the image is in the deprecated or deleted state.

guangyee (approved):
lets do this

PR #125, opened 2022-07-27
Use aws python3.9 image instead of python3.6

guangyee:
AWS python3.9 image uses Amazon Linux 2 [1], which supports
native psycopg2-binary (postgre driver) out-of-the-box.
We no longer need to copying binaries around as we may run into
the risk of GLIBC incompatibilities.


mbelur (approved):
LGTM

jeremy-moffitt (approved):
LGTM , caveat I did not test it. Much cleaner :)

PR #126, opened 2022-07-27
Trim images payload

guangyee:
AWS Lambda has a (non-configurable) maximum payload size of 6MB.
To ensure we stay under that limit, we will trim off the
older images, based on publishedon date, when necessary.

### DIFF
--- a/pint_server/app.py
+++ b/pint_server/app.py
@@ -16,13 +16,23 @@
 # you may find current contact information at www.suse.com
 
 import datetime
+import math
 import re
 from decimal import Decimal
-from flask import (abort, Flask, jsonify, make_response, request, redirect,
-                   Response)
+from flask import (
+    abort,
+    Flask,
+    jsonify,
+    make_response,
+    redirect,
+    request,
+    Response)
 from flask_cors import CORS
 from flask_sqlalchemy import SQLAlchemy
-from sqlalchemy import text, or_
+from sqlalchemy import (
+    desc,
+    or_,
+    text)
 from sqlalchemy.exc import DataError
 from sqlalchemy.orm.exc import NoResultFound, MultipleResultsFound
 from xml.dom import minidom
@@ -93,6 +103,10 @@ PROVIDER_SERVERS_EXCLUDE_ATTRS = {
 }
 
 SUPPORTED_CATEGORIES = ['images', 'servers']
+
+# NOTE: AWS lambda payload size cannot exceed 6MB. We are setting the
+# maximum payload size to 5.5MB to account for the HTTP protocol overheads
+MAX_PAYLOAD_SIZE = 5500000
 
 
 def get_supported_providers():
@@ -425,11 +439,38 @@ def get_provider_servers(provider):
             for server in servers]
 
 
+def trim_images_payload(images):
+    payload_size = jsonify(images=images).content_length
+
+    if payload_size >  MAX_PAYLOAD_SIZE:
+        # NOTE: assuming the size of the entries are evenly distributed, we
+        # determine the percentage of entries to trim from the end of the list,
+        # as the list is sorted in decending order by publishedon date, by
+        # calculating the percentage over the maximum payload size. Then
+        # trim the same percentage off the list, rounding up to be safe.
+        trim_size  = math.ceil(
+            ((payload_size - MAX_PAYLOAD_SIZE) / payload_size) * len(images))
+        last_publishedon = images[-trim_size]['publishedon']
+        images = images[:-trim_size]
+
+        # Now make sure we don't have partial data by finished triming all the
+        # images from all regions that have the same publishedon date that of
+        # the last image that got trimmed.
+        trim_size = 0
+        while images[-(trim_size + 1)]['publishedon'] == last_publishedon:
+            trim_size += 1
+        if trim_size:
+            images = images[:-trim_size]
+    return images
+
+
 def get_provider_images(provider):
-    images = PROVIDER_IMAGES_MODEL_MAP[provider].query.all()
+    images = PROVIDER_IMAGES_MODEL_MAP[provider].query.order_by(
+        desc(PROVIDER_IMAGES_MODEL_MAP[provider].publishedon)).all()
     exclude_attrs = PROVIDER_IMAGES_EXCLUDE_ATTRS.get(provider)
-    return [get_formatted_dict(image, exclude_attrs=exclude_attrs)
-            for image in images]
+    return trim_images_payload(
+            [get_formatted_dict(image, exclude_attrs=exclude_attrs)
+                for image in images])
 
 
 def get_data_version_for_provider_category(provider, category):


pint_server/app.py
```diff
@@ -425,11 +439,28 @@ def get_provider_servers(provider):
...
+
+    if payload_size >  MAX_PAYLOAD_SIZE:
+        # NOTE: assuming the size of the entries are evenly distributed, we
+        # determine the percentage of entries to trim from the end of the list,
+        # as the list is sorted in decending order by publishedon date, by
+        # calculating the percentage over the maximum payload size. Then
+        # trim the same percentage off the list, rounding up to be safe.
+        trim_size  = math.ceil(
+            ((payload_size - MAX_PAYLOAD_SIZE) / payload_size) * len(images))
+        images = images[:-trim_size]
```
rjschwei, line 453: This will create the problem that we might send partial data for a given image. For example we might send data for `suse-sles-12-sp1-v20170630-hvm...` for only some regions. Logic that is strictly based on math is not sufficient. We'll have to find the first image name that got cut and then make sure no images with that name are in the data that we would send.

rjschwei (changes requested):

pint_server/app.py
guangyee: Make sense. Will do.

pint_server/app.py
```diff
@@ -425,11 +439,35 @@ def get_provider_servers(provider):
...
+
+        # Now make sure we don't have partial data by finished triming all the
+        # images from all regions that have the same publishedon date that of
+        # the last image that got trimmed.
+        publishedon = images[-1]['publishedon']
+        while images[-1]['publishedon'] == publishedon:
```
rtamalin, line 459: Minor performance nit...

This while loop is going to lead to multiple `images` list modification operations because we pop off each extra entry individually, once per iteration.

If you instead just decrement the index value on each iteration, then do a single slice operation at the end to determine the final trimmed images list, that is likely to be slightly more efficient, performance wise.

Of course this is probably very minor in the overall scheme of things. ;-)

Similarly if there is a concern about avoiding too many list deref's, you could combine a computed index value with a step factor, e.g. go back in decrements of X (e.g. number of provider regions) and then use a binary search to find the first entry with the new publishedon value.

Again the performance gain would probably be negligible, unless we could potentially be dealing with many hundreds of images being published on the same day...

jeremy-moffitt: I agree on the pop... would be better to find the index and then make a single additional cut... I worry the binary search, which performant, may be overly complicated for the number of rows we're talking about... personally I'd probably keep it simple if possible.

rtamalin: Yeah I agree on binary search... Not sure we are anywhere near having 1000s of entries with same publishedon date...

pint_server/app.py
```diff
@@ -425,11 +439,35 @@ def get_provider_servers(provider):
...
+
+
 def get_provider_images(provider):
-    images = PROVIDER_IMAGES_MODEL_MAP[provider].query.all()
+    images = PROVIDER_IMAGES_MODEL_MAP[provider].query.order_by(
+        desc(PROVIDER_IMAGES_MODEL_MAP[provider].publishedon)).all()
     exclude_attrs = PROVIDER_IMAGES_EXCLUDE_ATTRS.get(provider)
-    return [get_formatted_dict(image, exclude_attrs=exclude_attrs)
-            for image in images]
+    return trim_images_payload(
```
rtamalin, line 468: So with this change all `/v1/<provider>/images` queries will be trimmed, no matter whether XML or JSON was requested... Is that OK?

pint_server/app.py
```diff
@@ -425,11 +439,35 @@ def get_provider_servers(provider):
             for server in servers]
 
 
+def trim_images_payload(images):
+    payload_size = jsonify(images=images).content_length
```
rtamalin, line 443: Note that this operation represents wasted work, since we call `jsonify()` just to get the content length here... And this will be wasted work for every `/v1/<provider>/images` request, whether the request was for JSON or XML...

If this is a concern we probably should restructure things such that a trim function (or set of functions) is passed as an argument to the `make_response()` routine, and it does the `jsonify()`, and checks the content length of that, and only if that exceeds the MAX_PAYLOAD_SIZE do we call the trim function to do any trimming...

That would allow for making a trimming mechanism more generic, and encoding specific, e.g. we wouldn't trim XML unless the XML encoded output exceeded the payload size limit.

pint_server/app.py
guangyee: no need for binary search as we don't have 1000s of entries to trim. We need to avoid binary search as performance is terrible.

pint_server/app.py
rtamalin: Per discussion in chat, this is intentional/desirable as we want the XML and JSON result to be equivalent.

pint_server/app.py
guangyee: Calling jsonify() give us an more accurate look at the actual payload size. It will be applicable to both XML and JSON payloads, as the XML payload is just a straight conversion from JSON.

rtamalin: Per discussion elsewhere, it is desirable that the XML and JSON results be equivalent, so we will always want to determine whether the JSON results (since they seem to grow faster than XML for equivalent content) wouldn't exceed the payload limit... So it would only be wasted work for JSON queries where we will `jsonify()` the images list twice, discarding this one after having retrieved the content_length...

guangyee: I can't imagine jsonify() add a noticeable overhead though.

pint_server/app.py
```diff
@@ -425,11 +439,39 @@ def get_provider_servers(provider):
...
+        trim_size  = math.ceil(
+            ((payload_size - MAX_PAYLOAD_SIZE) / payload_size) * len(images))
+        images = images[:-trim_size]
+
+        # Now make sure we don't have partial data by finished triming all the
+        # images from all regions that have the same publishedon date that of
+        # the last image that got trimmed. But if the last entry is already at
+        # the borard, this algorithm will trim off the entire data set with
+        # the same publishedon date, which is not a concern for us.
+        publishedon = images[-1]['publishedon']
```
rjschwei, line 460: The problem with this is that if the trim happens to land on the boundary then we trim images unnecessarily. The `publishedon` date that drives the trim really has to come from the first element of the slice that is getting trimmed off, i.e. 
```
last_publishedon = images[trim_size-1]['publishedon']
```
and this needs to be before

```
images = images[:-trim_size]
```

and then the while loop changes to

```
while images[-trim_size]['publishedon'] == last_publishedon
```
This avoids trimming a whole set of images if the trim happens to land on a "publishedon" boundary.

rjschwei (changes requested):

pint_server/app.py
guangyee: I've added the boundary scenario in the code comment, that trimming extra is not a concern. But I can update the code to account for the boundary scenario.

rjschwei (approved):

pint_server/app.py
rtamalin: Probably not in the case of relatively small data hierarchy, but walking the data structure and rendering it as a JSON string representation for a large data object does represent a body of work, given that in Python strings are immutable objects. The `jsonify()` is going to be a fairly optimised implementation, so the overhead will be minimised, but always something to keep in mind.

It is less of a concern since we have to do it anyway since we want the XML and JSON results to be equivalent, so we need to always `jsonfiy()` to get the content_length, so I don't really see a problem... I was just raising it initially as at first glance we were `jsonify()`ing the image list even for XML output requests, which seemed counter-intuitive until I realised that it may be desirable to ensure that the XML and JSON results would be equivalent.

As such I'm happy to resolve this discussion with no changes required.

pint_server/app.py
```diff
@@ -16,13 +16,23 @@
...
+import math
 import re
 from decimal import Decimal
-from flask import (abort, Flask, jsonify, make_response, request, redirect,
-                   Response)
+from flask import (
+    abort,
+    Flask,
+    jsonify,
+    make_response,
```
rtamalin, line 26: FYI just noticed that we are importing a make_response from flask here, but further down we define a make_response at line 501...

rtamalin (approved):
Looks good, and can be merged as is...

pint_server/app.py
jesusbv: It seems to me the import of `make_response` from Flask is not neeeded ?

jesusbv (approved):
other than the import of make_response, LGTM

pint_server/app.py
guangyee: good catch!

PR #127, opened 2022-07-29
Remove unnecessary import of make_response from flask

rtamalin:
A make_response routine is defined later in app.py, so there is no
need to import one from flask.


guangyee (approved):
Good catch!

PR #128, opened 2022-08-01
Fix unit test for DB version

guangyee:
Server is returning 'database server version: 14.2' so the test
should look for that instead of 'postgreSQL server version: 14.2'.


rtamalin (approved):
Looks good.

IIRC the switch from `postgreSQL` to just `database` in the response value was a late change during the review process

jeremy-moffitt (approved):
trivial fix, LGTM

PR #130, opened 2022-08-17
Don't include deleted images in /v1/{provider}/images response

rtamalin:
To reduce the payload size for the /v1/{provider}/images requests
we can just provide a response that includes the active, inactive
and deprecated images.

If someone wants the deleted images they can explicitly request them
with a /v1/{provider}/images/deleted request.

### DIFF
--- a/pint_server/app.py
+++ b/pint_server/app.py
@@ -691,7 +691,11 @@ def trim_images_payload(images):
 
 
 def get_provider_images(provider):
-    images = PROVIDER_IMAGES_MODEL_MAP[provider].query.order_by(
+    images = PROVIDER_IMAGES_MODEL_MAP[provider].query.filter(
+            PROVIDER_IMAGES_MODEL_MAP[provider].state.in_([
+                ImageState.active,
+                ImageState.inactive,
+                ImageState.deprecated])).order_by(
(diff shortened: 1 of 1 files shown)


pint_server/app.py
```diff
@@ -691,7 +691,11 @@ def trim_images_payload(images):
 
 
 def get_provider_images(provider):
-    images = PROVIDER_IMAGES_MODEL_MAP[provider].query.order_by(
+    images = PROVIDER_IMAGES_MODEL_MAP[provider].query.filter(
+            PROVIDER_IMAGES_MODEL_MAP[provider].state.in_([
```
guangyee, line 695: Maybe use `not_in() instead of `in()` as we only want to exclude the deleted ones? In case we add more states in the future.

rtamalin: Don't care either way ;-)

rjschwei (approved):

pint_server/app.py
rjschwei: I like the positive test, i.e. `in`

rtamalin: So just tested and there isn't a `not_in` operator available for the `state` field as it is... Using the `not_in()` would require additional plumbing to import it and structure the code to leverage it, so sticking with the `in_()` is the better approach for now.

guangyee (approved):
lets do this

PR #138, opened 2024-05-14
Truncate image for list images for a provider and a given state

guangyee:
To avoid going over the AWS lambda payload size, we need to truncate the payload when listing images for a provider and a given image state (i.e. deleted). The truncated images are based on the oldest deleted date.

Also update requirements.txt and test-requirements.txt with the latest available versions for Python 3.6.15.


pint_server/app.py
```diff
@@ -453,7 +453,8 @@ def get_provider_images_for_region_and_state(provider, region, state):
 def get_provider_images_for_state(provider, state):
     images = PROVIDER_IMAGES_MODEL_MAP[provider].query.filter(
         PROVIDER_IMAGES_MODEL_MAP[provider].state == state)
-    return formatted_provider_images(provider, images)
+    return trim_images_payload(
```
mbelur, line 456: to handle the deleted state images, do we need to use the deletedon attribute to sort and slice rather than the publishedon attribute?

test-requirements.txt
```diff
@@ -1,10 +1,10 @@
...
-mock
-Flask
-Flask-CORS
-Flask-SQLAlchemy
-lxml
-requests
-pytest-stress
-locust
-python-dateutil
+Flask==2.0.3
```
rtamalin, line 1: What do these versions represent? The latest available versions compatible with a specific Python version? The versions matching what is available in SLE for that Python version? Or something else...

It might be a good idea to provide a comment at the start of this file, and of the requirements.txt above, indicating the provenance of the versions being chosen.

pint_server/app.py
mbelur: thanks for the clarification.

test-requirements.txt
guangyee: That's the lowest common denominator essentially, that it is the highest version that Python 3.6.15 can support.

mbelur (approved):
LGTM!

rtamalin (approved):
LGTM

My concern about a lack of clarity as to how the requirements.txt and test-requirements.txt versions were selected has been addressed. As to whether that is the best approach in the long run is a separate discussion we can address after this PR has been merged. ;-)

PR #140, opened 2024-07-19
Sort all image queries by publishedon date

jeremy-moffitt:
To ensure that images trimmed from oversized queries are the oldest ones, it is necessary to sort the image/state query by publishedon date. For consistency, sort the other image queries the same way.

### DIFF
--- a/pint_server/app.py
+++ b/pint_server/app.py
@@ -32,6 +32,7 @@ from flask_cors import CORS
 from flask_sqlalchemy import SQLAlchemy
 from sqlalchemy import (
     desc,
+    asc,
     or_,
     text)
 from sqlalchemy.exc import DataError
@@ -408,17 +409,19 @@ def _get_azure_environment_name_for_region(region):
 
     return environment.environment
 
-def _get_azure_images_for_region_state(region, state=None):
+def _get_azure_images_for_region_state(provider, region, state=None):
     environment_name = _get_azure_environment_name_for_region(region)
 
     # query all images with matching environment and state (if specified)
     if state is None:
         images = MicrosoftImagesModel.query.filter(
-            MicrosoftImagesModel.environment == environment_name)
+            MicrosoftImagesModel.environment == environment_name).order_by(
+            desc(PROVIDER_IMAGES_MODEL_MAP[provider].publishedon)).all()
     else:
         images = MicrosoftImagesModel.query.filter(
             MicrosoftImagesModel.environment == environment_name,
-            MicrosoftImagesModel.state == state)
+            MicrosoftImagesModel.state == state).order_by(
+            desc(PROVIDER_IMAGES_MODEL_MAP[provider].publishedon)).all()
 
     return images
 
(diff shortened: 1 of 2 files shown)


rtamalin (approved):
LGTM

pint_server/app.py
```diff
@@ -414,11 +414,13 @@ def _get_azure_images_for_region_state(region, state=None):
     # query all images with matching environment and state (if specified)
     if state is None:
         images = MicrosoftImagesModel.query.filter(
-            MicrosoftImagesModel.environment == environment_name)
+            MicrosoftImagesModel.environment == environment_name).order_by(
+            desc(PROVIDER_IMAGES_MODEL_MAP[provider].publishedon)).all()
```
guangyee, line 418: `provider` is not defined in this function. Will need to pass it in.

pint_server/app.py
```diff
@@ -494,15 +500,15 @@ def query_deletedby_images_in_provider_region(
                 PROVIDER_IMAGES_MODEL_MAP[provider].region == region,
                 PROVIDER_IMAGES_MODEL_MAP[provider].state == ImageState.deprecated,
                 PROVIDER_IMAGES_MODEL_MAP[provider].deprecatedon < deprecatedby,
-            )
+            ).order_by(desc(PROVIDER_IMAGES_MODEL_MAP[provider].publishedon)).all()
```
guangyee, line 503: Since this is querying deleted images, should we sort by `deletedon` in ascending order?

pint_server/app.py
```diff
@@ -526,7 +532,8 @@ def _query_image_in_azure_region(image_name, region):
     # retrieve matching images for region
     images = MicrosoftImagesModel.query.filter(
         MicrosoftImagesModel.environment == environment_name,
-        MicrosoftImagesModel.name == image_name)
+        MicrosoftImagesModel.name == image_name).order_by(
+        desc(PROVIDER_IMAGES_MODEL_MAP[provider].publishedon)).all()
```
guangyee, line 536: same as above, `provider` must be passed in

pint_server/app.py
jeremy-moffitt: if we're trimming by the publishedon date, we should sort by the publishedon date

guangyee: We are not trimming this one, yet.

pint_server/app.py
```diff
@@ -466,7 +471,8 @@ def _get_azure_deprecatedby_images_for_region(deprecatedby, region):
         MicrosoftImagesModel.environment == environment_name,
         MicrosoftImagesModel.state == ImageState.deprecated,
         MicrosoftImagesModel.deprecatedon < deprecatedby,
-    )
+    ).order_by(desc(PROVIDER_IMAGES_MODEL_MAP[provider].publishedon)).all()
```
guangyee, line 474: We need to pass in `provider` into this function as well.

pint_server/app.py
```diff
@@ -494,15 +500,15 @@ def query_deletedby_images_in_provider_region(
                 PROVIDER_IMAGES_MODEL_MAP[provider].region == region,
                 PROVIDER_IMAGES_MODEL_MAP[provider].state == ImageState.deprecated,
                 PROVIDER_IMAGES_MODEL_MAP[provider].deprecatedon < deprecatedby,
-            )
+            ).order_by(asc(PROVIDER_IMAGES_MODEL_MAP[provider].deletedon)).all()
```
guangyee, line 503: `asc` need to be imported at the top.

pint_server/app.py
guangyee, line 1: Got this error on line 573 while doing API sanity tests. `TypeError: count() takes exactly one argument (0 given)`.
I think we need to change it to `if len(images) == 0:` since images is a list object.

jeremy-moffitt: made the recommended change, though this has me a little nervous, not because of the len check... but because I didn't anticipate this would change the type of the images collection...

rtamalin: This probably goes back to aggressively adding the .all() in all those previous cases.

Without the .all() the object returned is a list generator that and be used to render an actual list if needed, but it is also an SQLAlchemy query object that can have further query operations applied to it, including the count() method which would return a count of the number of entries in the query response. Once we actualised the query results into a Python list the count method is no longer allowed, and you need to use len() to obtain the length of the list.

PR #143, opened 2024-11-01
Migrate to pint_models package

smarlowucf:
Moving shared models and database utitilities to a separate package.


rtamalin:
Code looks good but checks are failing

rtamalin:
Looks like the pint-models package doesn't support older Python versions, and this package doesn't support newer ones ;-)

rtamalin (approved):
LGTM - but there are check pipeline issues to resolve ;-)

pint_server/app.py
```diff
@@ -42,8 +42,8 @@
 import xml.etree.ElementTree as ET
 
 import pint_server
-from pint_server.database import init_db, get_psql_server_version
-from pint_server.models import (ImageState, AmazonImagesModel,
+from pint_modesl.database import init_db, get_psql_server_version
```
rjschwei, line 45: s/pint_modesl/pint_models/

rjschwei (changes requested):

smarlowucf:
> Code looks good but checks are failing

Yep, after a bit of digging I realize that pyproject.toml wasn't supported in python 3.6. So the pint_models package needs to be reverted to setup.py until 3.6 support is dropped. Or pint server code drops support for python 3.6.

rjschwei:
I do not really see a reason why pint cannot/does not run on Python 3.11

jeremy-moffitt:
> I do not really see a reason why pint cannot/does not run on Python 3.11

we just need to test it and update the deployment... we've kicked the can a few times on this one, looks like it finally caught up with us :-/

rtamalin:
@rjschwei IIRC I have run it under Python 3.10 or 3.11, so the code itself should run fine, but the setup.py needs to be updated to reflect that, and we need to redo the container and associated package builds to use the python311 ecosystem; right now it is using the SLE 15 stream default python ecosystem, i.e. Python 3.6 :-(

rjschwei (approved):

rjschwei:
> I assume it would take some time to migrate to 3.11. Is it better to revert the public-cloud-info-models package to use setup.py temporarily until this can happen? That way the PR's aren't all stuck waiting?

That should be a matter of changing the spec file and setup.py, meaning probably not more than 1 hours worth of work.

PR #144, opened 2025-04-22
Pint-NG server compatible with Python 3.11 or later

chajain:


jeremy-moffitt (approved):
I had 2 github windows open and submitted my review in the wrong one on accident, ignore my initial review.

Dockerfile.sle15
```diff
@@ -1,4 +1,5 @@
-FROM registry.suse.de/devel/pubcloud/productionservices/pint/containers/pint/pint-server-lambda-function:latest
+#FROM registry.suse.de/devel/pubcloud/productionservices/pint/containers/pint/pint-server-lambda-function:latest
```
rtamalin, line 1: Potentially this could be handled with an ARG setting that defaults to the standard path, that can be overridden via a command line argument when building the image to point to the alternate path.

However the value landed by the PR should use the standard path, not a developer's personal path.

rtamalin (changes requested):
The PR submission should use standard paths, not a developers personal path.

jeremy-moffitt (commented):
Question about these changes in general... it seems we're still hard-coding the version, just a newer version. Is it possible to get away from that? With this approach it seems we'll have to repeat this process as on-going maintenance.

rtamalin:
> Question about these changes in general... it seems we're still hard-coding the version, just a newer version. Is it possible to get away from that? With this approach it seems we'll have to repeat this process as on-going maintenance.

Agreed, but I think we should get these changes in so that we have a 3.11 based version, and then work on the decoupling aspect of things.

.github/workflows/ci-workflow.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        python-version: [3.6.15, 3.7, 3.8]
+        python-version: [3.11]
```
rtamalin, line 13: Note that since these tests are running independently of the target SLE deployment ecosystem, there is no hard in having them run for multiple viable versions of Python.

However, Python 3.6 and Python 3.7 are not viable within the ubuntu-22.04 and later runners, so the minimum version that you can test for is 3.8.

And per my #145 the package versions in the requirements.txt and/or test-requirements.txt are not compatible with Python 3.12 or later, so this list should be left at 3.8, 3.9, "3.10", and 3.11. Note the quoting around "3.10" because without it the value will be treated as a floating point number by YAML parsing and 3.10 == 3.1.

setup.cfg
```diff
@@ -15,8 +15,8 @@ classifier =
     Programming Language :: Python :: Implementation :: CPython
     Programming Language :: Python :: 3 :: Only
     Programming Language :: Python :: 3
-    Programming Language :: Python :: 3.6
+    Programming Language :: Python :: 3.11
```
rtamalin, line 18: Technically we should be listing all applicable Python versions here, not just the target SLE deployment version, so we should really have entries for 3.8 through 3.11, and Python 3.6 and 3.7 entries are still valid, if outdated.

rtamalin:
Something we should consider is putting the spec file used to build the pint server package into this repo, and then updating the package's _service file in the build service to populate the spec file from the repo as well.

This would help keep all of the relevant files for Python version management in a single place, and it is primarily in the spec file that we will need to do the work to make things version agnostic.

Dockerfile.sle15
```diff
@@ -1,4 +1,4 @@
-FROM registry.suse.de/devel/pubcloud/productionservices/pint/containers/pint/pint-server-lambda-function:latest
+FROM registry.suse.de/home/chajain/devel/pubcloud/productionservices/pint/containers/pint/pint-server-lambda-function:latest
```
rtamalin, line 1: You still have this using your personal build area path, rather than the standard build area path, as should be the case in a change that is being submitted to the code base.

rtamalin (changes requested):
Looks good except for the lingering reference to the home/chajain project space in the Dockerfile.sle15 FROM line.

Dockerfile.sle15
chajain: Sorry. I thought I fixed it. Let me do that again

rtamalin (approved):
LGTM

jeremy-moffitt (approved):

PR #145, opened 2025-04-23
Update GitHub Actions CI workflow to use ubuntu-latest

rtamalin:
Update the standard CI workflow job definition to target the latest ubuntu runner type.

Rename CI workflow jobs to reflect that they are running tests rather than building.

Create a duplicate job definition for legacy python versions that targets ubuntu-20.04, that is disabled under GitHub Actions runs, but still available to be tested locallly using nektos/act or the gh act extension.

Note that for versions of Python greater that 3.11 testing fails due to incompatible requirements.txt and test-requirements.txt package versions; these versions have been disabled until that issue has been addressed.

Ensure that the PATH is setup appropriately when using runners that run as a non-root user.

Update checkout and setup-python action versions to latest.

Ensure that the dev and test venv creation helper scripts install latest applicable versions of the pip, wheel and setuptools packages.

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -5,24 +5,68 @@ on:
   - pull_request 
 
 jobs:
-  build:
 
+  # test for versions that require an older distro
+  test-legacy-versions:
+
+    # will be skipped in GitHub infrastructure, but can be run locally
+    # using nektos/act if an appropriate image is available and the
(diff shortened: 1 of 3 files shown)


jeremy-moffitt (approved):
LGTM, but should get a more technical reviewer as the primary

chajain (commented):
Looks good

PR #147, opened 2025-06-09
PINT APIs for querying data in a compressed manner

chajain:
Implementing bumpversion
Adding python-pintserver.spec file under package folder
Update the package versions in requirements.txt  so that it works for Python 3.12 and 3.12+
Add new APIs so that data for each API can be sent across the network in a compressed fashion,


package/python-PintServer.changes
```diff
@@ -0,0 +1,77 @@
+-------------------------------------------------------------------
+Tue Apr 29 18:59:03 UTC 2025 - Charu Jain <cjain@suse.com>
+
+- Update to 2.0.16
```
rtamalin, line 4: Again, shouldn't you have a 2.0.17 version entry?

pint_server/app.py
```diff
@@ -775,12 +777,21 @@ def make_response(content_dict, collection_name, element_name):
             content = {collection_name: content_dict}
         else:
             content = content_dict
-        return jsonify(**content)
+        if request.path.endswith('.gz'):
+            if request.path.endswith('.xml.gz'):
+               return Response( gzip.compress((json_to_xml(content_dict, collection_name, element_name)).encode('utf-8'), compresslevel=9), mimetype='application/gzip; charset=utf-8')
+            else:
+               return Response( gzip.compress((json.dumps(content, separators=(',', ':'))).encode('utf-8'), compresslevel=9), mimetype='application/gzip; charset=utf-8')
```
rtamalin, line 784: Suggest wrapping this long line also.

pint_server/app.py
```diff
@@ -775,12 +777,21 @@ def make_response(content_dict, collection_name, element_name):
             content = {collection_name: content_dict}
         else:
             content = content_dict
-        return jsonify(**content)
+        if request.path.endswith('.gz'):
+            if request.path.endswith('.xml.gz'):
```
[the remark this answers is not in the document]
rtamalin: Potentially you could restructure the make_response() to just have a since
```return Response()```
line at the end, and just have logic before hand to setup the two arguments...

Something like:
```
# generate xml or json formatted payload
if requested response format is xml
  payload = json_to_xml(content_dict, collection_name, element_name)
  app_type = 'xml'
else:
  if collection_name specified:
    content = {collection_name: content_dict}
  else
    content = content_dict
  payload = json.dumps(content, separators=(',', ':'))
  app_type = 'json'

# encode the payload with the desired charset
charset = 'utf-8'
payload = payload.encode(charset)

# if compressed response requested, compress the payload
if requested response is gzip
  payload = gzip.compress(payload, compresslevel=9)
  app_type = 'gzip'

# generate an appropriate mimetype value
mimetype = 'application/%s; charset=%s' % (app_type, charset)

# return the response
return Response(payload, mimetype=mimetype)
```

File: package/_service [hunk 4]
```diff
@@ -0,0 +1,14 @@
+<services>
+  <service name="tar_scm">
+    <param name="url">https://github.com/SUSE-Enceladus/public-cloud-info-service.git</param>
+    <param name="scm">git</param>
+    <param name="version">2.0.16</param>
+    <param name="revision">refs/tags/v2.0.16</param>
+    <param name="filename">pint-server</param>
+  </service>
```
rtamalin, line 8: Note that tar_scm has params that allow it to generate a .changes file entry based upon git commits since the previous version...

chajain:
Adding another PR

package/_service
smarlowucf, line 1: Do we want to have the service file in git? It could be part of the obs package. For example https://build.opensuse.org/projects/Cloud:Tools:CI/packages/mash-client/files/_service?expand=1. That automagically pulls the version from the tag and it's triggered by github to run and update the package when a new tag is created.

chajain:
Yes, service file is not part of git. That was an old PR which I have
closed. Please look at latest PR
https://github.com/SUSE-Enceladus/public-cloud-info-service/pull/150

Sorry for all the confusion

PR #149, opened 2025-06-25
Gzip compress

chajain:


package/python-PintServer.spec
```diff
@@ -24,8 +24,13 @@
 %global _sitelibdir %{%{pythons}_sitelib}
 
 Name:           python-PintServer
+<<<<<<< gzip_compress
```
brett060102, line 27: You missed this merge conflict.

PR #150, opened 2025-06-25
Querying data in compressed manner

chajain:
Provide new public v1 endpoints to deliver the data as a compressed stream

Remove use of "Germany Central" from functional test_get_microsoft_region_servers() as it is no longer a valid region name.

### DIFF
--- a/pint_server/tests/functional/test_function_app.py
+++ b/pint_server/tests/functional/test_function_app.py
@@ -18,6 +18,8 @@
 import json
 import pytest
 import requests
+import gzip
+import io
 
 
 # For some providers we need to remap the server type returned by
@@ -84,6 +86,23 @@ def _get_provider_region_images_in_state(baseurl, provider, region, state):
     return [i["name"] for i in images_resp.json()['images']]
 
 
+def _decompress_gzip(resp):
+    with gzip.GzipFile(fileobj=io.BytesIO(resp.content), mode='rb') as f:
+        uncompressed_data = f.read()
+    return uncompressed_data
+
+def _get_response_content(resp, extension):
+    if '.gz' in extension:
+        resp_content = _decompress_gzip(resp).decode('utf-8')
+        if '.xml' in extension:
+            assert '<?xml version=' in resp_content
+        else:
+            assert '{' in resp_content
+    else:
+        resp_content = resp.text
+    return resp_content    
+
+
 #
 # Helper functions for validating the data
 #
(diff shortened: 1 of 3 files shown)


pint_server/tests/functional/test_function_app.py
```diff
@@ -255,22 +285,25 @@ def test_get_provider_region_category(baseurl, provider, category, extension):
         # Pick the first region.. no need to iterate with all the regions
         region = regions[0]
         url = baseurl + '/v1/' + provider + '/' + region + '/' + category + extension
+        print(url)
```
rtamalin, line 288: Is this a left over debug statement?

pint_server/tests/functional/test_function_app.py
```diff
@@ -255,22 +285,25 @@ def test_get_provider_region_category(baseurl, provider, category, extension):
...
+@pytest.mark.parametrize("extension", ['', '.json', '.xml', '.gz', '.xml.gz', '.json.gz'])
 @pytest.mark.parametrize("category", ['images', 'servers'])
 def test_get_microsoft_region_servers(baseurl, category, extension):
     microsofttestregions = ['australiacentral', 'australiacentral2',
                             'Brazil South', 'brazilsouth',
                             'canadaeast', 'centralus', 'centraluseuap','chinaeast',
                             'East US', 'eastus',
                             'francecentral', 'francesouth',
-                            'Germany Central', 'germanycentral',
+                            'germanycentral',
```
rtamalin, line 306: How come you dropped `Germany Central`?

rtamalin (commented):
Looks very good... Have a couple of minor questions to be resolved, and happy to approve then.

rjschwei (changes requested):
I do not like this approach of adding new routes. We should use the standard way of indicating compression established in the protocol, i.e. the client sends a header that contains `Accept-Encoding` and then the server compresses according to the request. We can limit the compression formats to xz, bzip, gzip. When the server returns the data it sends `Content-Encoding` header and the client would then use the appropriate tool client side to decompress the data.

rtamalin:
> I do not like this approach of adding new routes. We should use the standard way of indicating compression established in the protocol, i.e. the client sends a header that contains `Accept-Encoding` and then the server compresses according to the request. We can limit the compression formats to xz, bzip, gzip. When the server returns the data it sends `Content-Encoding` header and the client would then use the appropriate tool client side to decompress the data.

Do we know if our CloudFront is configured to compress objects? Per [CloudFront compression docs](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/ServingCompressedFiles.html#compressed-content-cloudfront-notes) it may limit us to gzip or br (brotli) compression as other encoding types are stripped from the headers...

PR #151, opened 2025-07-02
Compression using the Accept-Encoding header of client Request

chajain:
Check Accept-Encoding header in the request sent by Client and compress the data depending on compression_type and Set Content_encoding to the compression_type used to compress


chajain:
Updated the original PR so that there is no confusion

PR #152, opened 2025-07-02
Querying data in compressed manner using gzip Encoding in request Header

chajain:
Compressing the data using Accept-Encoding in requests coming from Client

### DIFF
--- a/pint_server/app.py
+++ b/pint_server/app.py
@@ -19,6 +19,10 @@ import datetime
 import math
 import os
 import re
+import gzip
+import bz2
+import lzma
+import json 
 from collections import namedtuple
 from dateutil.relativedelta import relativedelta
 from decimal import Decimal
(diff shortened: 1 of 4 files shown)


brett060102 (approved):

rtamalin (changes requested):
We have a fall through assumption of lzma as the compression encoding to use if Accept-Encoding is set, even if lzma is not one of the specified values.

Also Accept-Encoding can have multiple comma separated values, possibly specifying client preference weightings, so we should handling that case properly. See https://www.rfc-editor.org/rfc/rfc9110.html#name-accept-encoding

pint_server/app.py
```diff
@@ -765,17 +774,110 @@ def assert_valid_date(date):
...
+    # encode the payload with the desired charset
+    payload = payload.encode(charset)
+    return app_type, payload
+
+
+def make_response(content_dict, collection_name, element_name):
+
+    charset = 'utf-8'
+    content_encoding_header = None
+    accept_encoding = None
```
rtamalin, line 805: You can initialise accept_encoding here and avoid having to fetch it twice below

```suggestion
    accept_encoding = request.headers.get('Accept-Encoding')
```

pint_server/app.py
```diff
@@ -765,17 +774,110 @@ def assert_valid_date(date):
...
+    app_type, payload = make_response_payload(
+        content_dict, collection_name, element_name, charset)
+    content_type = 'application/%s;charset=%s' % (app_type, charset)
+
+    if request.headers.get('Accept-Encoding') != None:
+        accept_encoding = request.headers.get('Accept-Encoding').split(", ")
```
rtamalin, line 811: Here you are splitting the value of the accept_encoding header on `, ` but above on line 698 & 699 you don't split it, and pass it as the string value.

In both cases the `in` check in the comprehension argument to the any() call in supported_encoding() will work, but in this case it will only match for exact matches, without risk of substring partial matches, because you are checking for the existence of an entry in a list or tuple... However, in the earlier case it is still susceptible to substring partial matches, since it is just checking for a substring being present in a string.

Finally the standard allows for the whitespace around the comma being optional, so you may be better off doing a split on `,` and then apply a strip() to the resulting list elements:

```suggestion
        accepted_encodings = [e.strip() for e in accept_encoding.split(",")]
```

rtamalin (changes requested):
Almost there, having added the fix to not assume lzma if the bzip2 and gzip2 checks didn't match.

However, the use of supported_encodings is inconsistent, with it being passed a string value in one case, meaning the internal `in` operator is doing a substring match checking for our supported encoding types, but in the other case it is being passed a list of string values which means the `in` operator is doing a check for element existence in the list, which is the preferred approach.

rtamalin (approved):
LGTM

rtamalin (approved):
Still LGTM

